Issue #1, opened 2017-03-01
Widget Organization & Class Structure


celloexpressions:
Since all media controls will have a shared set of code (and UI) to handle the media modal and the widget frame (title, caption, filters, etc.), there should be a parent class with the base functionality and child classes for the specific widgets. This would largely echo the way that customize controls work. I'm not sure if this is currently possible with widgets, but it would seem like a prerequisite for building separate image/video/audio/document widgets to me. Duplicating that baseline media code seems like a no-go.

I also wonder whether building these as separate widgets could actually cause more issues in a block-based system, where these widgets have to be merged with shortcodes and new blocks if all of these things come together. That could cause major backwards-compatibility issues within the next few months. This also relates to a comment I brought up in the context of editor blocks: when adding a media-type block, any type of media would likely bring up the existing media modal, at which point a more specific type of media could be selected. Or, even better, you could upload (or select) any type of file, then when you insert it it becomes a more specific type of block automatically. Rather than having to select the exact type of content from a much larger list of blocks, it's done automatically, which becomes particularly useful once you get into types of files that we may not be thinking about, starting with PDFs or .zips and extending to a very broad array of potential file types. It also seems to align with the core philosophies of striving for simplicity and decisions, not options. That approach can and should be balanced with discoverability but I doubt that making users select very-specific block types from the beginning is the best approach (and that doesn't necessarily mean that there couldn't also be a distinct image block that's tied into the more generic functionality - these are all questions that need to be researched).

All of that to say, if we want to launch something in the very near future as a quick win, we should go back to the more flexible media widget case, and constrain functionality where it makes sense for now. If we're looking for a more forward-thinking or longer-term approach, we should probably work out the best actual block-based solution rather than coming up with a set of interim widgets that could cause more problems than they solve.

westonruter:
> Since all media controls will have a shared set of code (and UI) to handle the media modal and the widget frame (title, caption, filters, etc.), there should be a parent class with the base functionality and child classes for the specific widgets. This would largely echo the way that customize controls work. I'm not sure if this is currently possible with widgets, but it would seem like a prerequisite for building separate image/video/audio/document widgets to me. Duplicating that baseline media code seems like a no-go.

If there is enough duplicated code, this can be done by implementing a `abstract` class at the PHP level that the different media type widgets inherit from.

> I also wonder whether building these as separate widgets could actually cause more issues in a block-based system, where these widgets have to be merged with shortcodes and new blocks if all of these things come together.

[...]

PR #2, opened 2017-03-08
Offer a dedicated widget per media type.

obenland:
This PR separates the media widgets into three media-type specific widgets.

See #1.


westonruter (approved):
Only defect I see is the selective refresh of a video or audio player fails to re-initialize ME.js, but this is to be expected.

PR #4, opened 2017-03-08
Misc cleanup for media widget abstract class

westonruter:


wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -53,8 +53,9 @@ public function __construct( $id_base = '', $name = '', $widget_options = array(
 		$control_opts = wp_parse_args( $control_options, array() );
 
 		parent::__construct(
-			$id_base ? $id_base : 'wp-media-widget', // @todo This should just be 'media'.
-			$name ? $name : __( 'Media' ),
+			$id_base,
+			/* translators: placeholder is the id base of the widget */
+			$name ? $name : sprintf( __( 'Media: %s' ), $id_base ),
```
obenland: So the media type would be lowercase and untranslatable? I think I'd prefer to either not make `$id_base` and `$name` optional instead

obenland (changes requested):

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -54,8 +54,7 @@ public function __construct( $id_base = '', $name = '', $widget_options = array(
 
 		parent::__construct(
 			$id_base,
-			/* translators: placeholder is the id base of the widget */
-			$name ? $name : sprintf( __( 'Media: %s' ), $id_base ),
+			$name,
```
obenland: Would you want to go as far as removing the empty string fallbacks in the method's arguments list?

westonruter: I think that makes sense, yes 👍

obenland (approved):

PR #5, opened 2017-03-08
Add initial support for re-initializing media elements upon selective refresh

westonruter:
This fixes an issue with the video and audio widgets where a change to the widget in the customizer causes the widgets to selectively refresh and then fail to re-initialize ME.js.

### DIFF
--- a/core-media-widgets.php
+++ b/core-media-widgets.php
@@ -31,6 +31,8 @@
  */
 function wp32417_default_scripts( WP_Scripts $scripts ) {
 	$scripts->add( 'wp-media-widget', plugin_dir_url( __FILE__ ) . 'wp-admin/js/widgets/media-widget.js', array( 'jquery', 'media-models', 'media-views', 'wp-mediaelement' ) );
+
+	$scripts->add_inline_script( 'customize-selective-refresh', file_get_contents( dirname( __FILE__ ) . '/wp-includes/js/customize-selective-refresh-extras.js' ) );
 }
 add_action( 'wp_default_scripts', 'wp32417_default_scripts' );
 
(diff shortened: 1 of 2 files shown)


obenland (approved):
The `file_get_contents` feels icky, but it's only temporary, seems fine

westonruter:
@obenland I agree that using the inline script feeling icky. What are your thoughts on the todo?

> If audio, first ensure that wp_audio_shortcode_library filters away mediaelement; if video, check wp_video_shortcode_library filters the same

I've never seem this in the wild, but let's say someone adds a plugin that does:

```php
add_filter( 'wp_video_shortcode_library', function(){ return 'some_other_library'; } );
add_filter( 'wp_audio_shortcode_library', function(){ return 'some_other_library'; } );
```

Now, this should result in MediaElement.js not being loaded. Maybe this is out of scope for the media widgets, but if I only override _one_ of the libraries then MediaElement.js still ends up initializing both of them since it uses this selector: `$( '.wp-audio-shortcode, .wp-video-shortcode' )`.

obenland:
Yeah, I don't think that is something we should concern ourselves with in the context of this plugin.

Not sure how to best go about it either, truthfully

westonruter:
@joemcgill Is `wp.mediaelement` wrong to do `$( '.wp-audio-shortcode, .wp-video-shortcode' )` if a user had filtered `wp_video_shortcode_library` or `wp_audio_shortcode_library` to not be `mediaelement`? It seems maybe the values for these filters should be exported to `_wpmejsSettings` and `wp.mediaelement.initialize()` should read from them when deciding whether `.wp-audio-shortcode` and/or `.wp-video-shortcode` should be included among the selector?

joemcgill:
@westonruter I'm not familiar with the back-history for that filter, but your thinking makes sense to me. I think it would have been a fairly safe assumption that if you were filtering the MJS library that you were doing so globally, and not optionally within the same page/frame. A workaround would be to make sure the already initialized frame had the `.mejs-mediaelement` class on it or we could alternately create a generic `.mediaelement--loaded` class that would be looked for before reinitializing.

PR #10, opened 2017-03-09
Prefix media widgets

obenland:
Prefixes the widgets `id_base` to not clash with existing audio/image/video widgets. Also adds a proper widget option to pass on the widgets mime type.

See https://wordpress.slack.com/archives/core-customize/p1489080794278492

### DIFF
--- a/wp-includes/widgets/class-wp-widget-audio.php
+++ b/wp-includes/widgets/class-wp-widget-audio.php
@@ -23,9 +23,10 @@ class WP_Widget_Audio extends WP_Widget_Media {
 	 * @access public
 	 */
 	public function __construct() {
-		parent::__construct( 'audio', __( 'Audio' ), array(
-			'classname'   => 'widget_audio',
+		parent::__construct( 'wp-audio', __( 'Audio' ), array(
+			'classname'   => 'widget_wp-audio',
 			'description' => __( 'Displays an audio file.' ),
+			'mime_type'   => 'audio',
 		) );
 
 		if ( $this->is_preview() ) {
--- a/wp-includes/widgets/class-wp-widget-audio.php
+++ b/wp-includes/widgets/class-wp-widget-audio.php
@@ -23,8 +23,8 @@ class WP_Widget_Audio extends WP_Widget_Media {
 	 * @access public
 	 */
 	public function __construct() {
-		parent::__construct( 'wp-audio', __( 'Audio' ), array(
-			'classname'   => 'widget_wp-audio',
+		parent::__construct( 'media-audio', __( 'Audio' ), array(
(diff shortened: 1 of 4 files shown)


wp-includes/widgets/class-wp-widget-audio.php
```diff
@@ -23,9 +23,10 @@ class WP_Widget_Audio extends WP_Widget_Media {
 	 * @access public
 	 */
 	public function __construct() {
-		parent::__construct( 'audio', __( 'Audio' ), array(
-			'classname'   => 'widget_audio',
+		parent::__construct( 'wp-audio', __( 'Audio' ), array(
```
westonruter: Thanks for the link to the [Slack conversation](https://wordpress.slack.com/archives/core-customize/p1489080794278492). It's not precedented to prefix widget id bases with `wp-` in core. Existing ID bases have `text`, `meta`, `search`, and so on. What about using `media_` instead as a prefix? This would seem more in line with existing core conventions and it would reflect that these are all subclasses of the one media widget.

Alternatively, as was done with term meta, existing plugins be updated to be compatible with the new widget as opposed to accommodating core for plugins that happen to use that ID base. This would be more difficult than using a prefix, however. I like `media_*`.

westonruter (changes requested):

wp-includes/widgets/class-wp-widget-audio.php
obenland: > What about using `media_` instead as a prefix?

I like that

wp-includes/widgets/class-wp-widget-audio.php
```diff
@@ -23,9 +23,10 @@ class WP_Widget_Audio extends WP_Widget_Media {
 	 * @access public
 	 */
 	public function __construct() {
-		parent::__construct( 'audio', __( 'Audio' ), array(
-			'classname'   => 'widget_audio',
+		parent::__construct( 'media-audio', __( 'Audio' ), array(
```
westonruter: One more thing…™

Should the ID base be `media_audio` or `media-audio`? Core is not consistent here. For multi-word widget ID bases, there are:

* `nav_menu`
* `recent-comments`
* `recent-posts`
* `tag_cloud`

So should we use `-` or `_`?

If I recall correctly, the REST API favors hyphens so I've inquired about that: https://wordpress.slack.com/archives/core-restapi/p1489088798492976

obenland: We should probably use `_`s :(

https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#naming-conventions

westonruter (approved):

PR #12, opened 2017-03-10
Standardize alignment handling

obenland:
* Fixes some bugs in Appearance -> Widgets
* Improves settings validation
* Uses `$default_instance` where needed.
* Moves some styles out of JS

### DIFF
--- a/wp-includes/widgets/class-wp-widget-image.php
+++ b/wp-includes/widgets/class-wp-widget-image.php
@@ -42,7 +42,9 @@ class WP_Widget_Image extends WP_Widget_Media {
 	 * @return void
 	 */
 	public function render_media( $attachment, $widget_id, $instance ) {
-		$instance    = wp_parse_args( $instance, array( 'size' => 'thumbnail' ) );
+		$instance = wp_parse_args( $instance, array(
+			'size' => 'thumbnail',
+		) );
 		$has_caption = ! empty( $attachment->post_excerpt );
 
 		$image_attributes = array(
@@ -53,7 +55,7 @@ class WP_Widget_Image extends WP_Widget_Media {
 		);
 
 		if ( ! $has_caption ) {
-			$image_attributes['class'] .= ' ' . $instance['align'];
+			$image_attributes['class'] .= ' align' . $instance['align'];
 		}
 
 		$image = wp_get_attachment_image( $attachment->ID, $instance['size'], false, $image_attributes );
@@ -71,12 +73,15 @@ class WP_Widget_Image extends WP_Widget_Media {
 		}
 
 		if ( $has_caption ) {
-			// [0] => width, [1] => height;
+			$width = 0;
 			$size = _wp_get_image_size_from_meta( $instance['size'], wp_get_attachment_metadata( $attachment->ID ) );
+			if ( false !== $size ) {
+				$width = $size[0];
+			}
 			$image = img_caption_shortcode( array(
 				'id'      => $widget_id . '-caption',
-				'width'   => isset( $size[0] ) ? $size[0] : false,
-				'align'   => $instance['align'],
+				'width'   => $width,
+				'align'   => 'align' . $instance['align'],
 				'caption' => $attachment->post_excerpt,
 			), $image );
 		}
--- a/wp-includes/widgets/class-wp-widget-image.php
+++ b/wp-includes/widgets/class-wp-widget-image.php
@@ -42,25 +42,28 @@ class WP_Widget_Image extends WP_Widget_Media {
 	 * @return void
 	 */
 	public function render_media( $attachment, $widget_id, $instance ) {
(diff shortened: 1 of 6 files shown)


wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -101,20 +102,25 @@ public function widget( $args, $instance ) {
...
 		// ID and title.
 		$instance['id']    = (int) $new_instance['id'];
 		$instance['title'] = sanitize_text_field( $new_instance['title'] );
 
-		// Everything else.
-		$instance['align'] = sanitize_text_field( $new_instance['align'] );
-		$instance['size']  = sanitize_text_field( $new_instance['size'] );
-		$instance['link']  = sanitize_text_field( $new_instance['link'] );
+		if ( in_array( $new_instance['align'], array( 'none', 'left', 'right', 'center' ) ) ) {
+			$instance['align'] = 'align' . $new_instance['align'];
```
westonruter: Should we normalize the `align` instance property to not its value also have an `align` prefix? We can prepend any value at the moment of rendering if the renderer needs it.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -101,20 +102,25 @@ public function widget( $args, $instance ) {
...
+		}
+
+		$image_sizes = array_merge( get_intermediate_image_sizes(), array( 'full' ) );
+		if ( in_array( $new_instance['size'], $image_sizes ) ) {
```
westonruter: Add `$strict = true` param for good measure.

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -42,25 +42,28 @@ public function __construct() {
...
 		$image = wp_get_attachment_image( $attachment->ID, $instance['size'], false, $image_attributes );
 
 		if ( $has_caption ) {
+			// [0] => width, [1] => height;
+			$size = _wp_get_image_size_from_meta( $instance['size'], wp_get_attachment_metadata( $attachment->ID ) );
 			$image = img_caption_shortcode( array(
 				'id'      => $widget_id . '-caption',
-				'width'   => get_option( $instance['size'] . '_size_w' ),
+				'width'   => isset( $size[0] ) ? $size[0] : false,
 				'align'   => $instance['align'],
```
westonruter: Per below, I think we can make this:

```php
'align' => 'align' . $instance['align'],
```

As this is an idiosyncrasy of `img_caption_shortcode` that we needn't canonize as part of the widget instance properties.

westonruter (changes requested):

wp-includes/widgets/class-wp-widget-media.php
obenland: I'm on the fence about it. We currently don't use if for anything other than that CSS class, which is why I thought we could just handle it in sanitization

westonruter: Potentially a custom widget renderer could want to use the value in a `float` CSS property.

obenland: `.align*` does exactly that. Better even because themes have optimized styles for it.

westonruter: Right, but I was just trying to think of an example. Like, when we expose these widgets in the REST API, I'd surely hope we'd be exposing the data like:

```json
{"align": "left"}
```

Instead of:

```json
{"align": "alignleft"}
```

Also, more importantly, in the sanitization logic here:

```php
if ( in_array( $new_instance['align'], array( 'none', 'left', 'right', 'center' ) ) )
```

This means that a sanitized `align` instance value of `alignleft` would itself fail this sanitization check, as `alignleft` would not pass the `in_array` check, and it wouldn't be possible to take an existing instance and say copy it onto another widget.

westonruter: To be clear, we'd still convert `left` to `alignleft` but only at the point we pass the value into `wp_get_attachment_image()`. The underlying saved widget property would be a “pristine” plain `left`.

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -42,17 +42,18 @@ public function __construct() {
...
 			'data-id' => $widget_id,
 			'title'   => $attachment->post_title,
 			'class'   => 'image wp-image-' . $attachment->ID,
-			'style'   => 'width: 100%; height: auto;',
+			'style'   => 'max-width: 100%; height: auto;',
 		);
 
 		if ( ! $has_caption ) {
-			$image_attributes['class'] .= ' align' . $instance['align'];
+			$image_attributes['class'] .= ' ' . $instance['align'];
```
westonruter: So I think this change should be reverted, so we continue to prefix the alignment property with `align` only here when output as the prefixing of `align` is really only relevant in the context of a CSS classname.

westonruter (approved):

PR #11, opened 2017-03-10
Add edit media button

adamsilverstein:

PR #14, opened 2017-03-10
Add edit media button #2

obenland:
See #11

### DIFF
--- a/wp-admin/css/widgets/media-widget.css
+++ b/wp-admin/css/widgets/media-widget.css
@@ -20,6 +20,12 @@
 	text-align: center
 }
 
+.media-widget-admin-preview:after {
+	content: '';
+	display: table;
+	clear: both;
+}
+
 .media-widget-buttons {
 	display: -webkit-box;
 	display: -moz-box;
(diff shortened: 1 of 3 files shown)


obenland:
This is not the way I was hoping it would turn out. :(

@adamsilverstein Would you mind if I added you as a collaborator and we continue working on this branch?

obenland:
I just added you, I figured you could decline the invite if you didn't want to ;)

obenland:
@westonruter To be able to remove captions in the control preview we'll need some other way to store the caption information, likely a hidden input. Given that we currently use the same function to render the component preview and the actual widget content, I'd suggest waiting with removing the caption visually until #17 is in and we have templates we can use?

westonruter:
@obenland yes, I agree. The caption should be removed from the control preview. The caption can be rendered exclusively in the customizer preview frame. And yes, the `caption` and other fields need to be stored in hidden inputs, and an initial implementation of this is now in #17 and ready for initial review. Some more cleanup is needed. Also, it's focused on the image widget. The audio and video widgets are currently disabled.

PR #17, opened 2017-03-12
[WIP] Widget JS Refactor

### LINKED ISSUE #7: Address flash of replaced content in widgets admin screen
On the widgets admin screen, saving a widget causes the new form's contents to be sent back from the server and they replace the previous form, followed by invoking `widget-updated` event. Since there are JS components in the media widget which need to be re-initialized, and the result here is the Media Element being destroyed and rebuilt each time you save the widget and there is a flash of replaced/uninitialized content, with prior to save.

This is a relatively small issue, but it is an annoyance. Widgets in the customizer don't do this same form replacement behavior if the widget's `form` callback returns the same inputs as already exist in the DOM.

A more modern way to integrate JS components into widgets is explored in the [JS Widgets](https://github.com/xwp/wp-js-widgets) feature plugin. Nevertheless, the JS Widgets plugin is too large in scope for a minor release to get media widgets out for users sooner rather than later, but at least I'd hope that we can keep it in mind.

----

For reference, given a widget that looks like this:

image

For the moment immediately after saving it will look like this:

image

### LINKED ISSUE #8: Refactor JS to use an extensible Backbone View
The current `wp.MediaWidget` is used primarily as a namespace for methods that are event handlers for the `widget-added` and `widget-updated` events. I think we can clean this up by creating a base Backbone View for representing a Media widget which can then be extended for the various media types being represented. Then, similarly to how the customizer has `wp.customize.controlConstructor` there can be a lookup of media type (widget `id_base`) to the specific View subclass that is responsible for managing the state of the media widget form.

### LINKED ISSUE #16: Alt and caption text are not shown
Just a reminder that  alt or caption text aren't hooked up yet.

1. Go to the Customizer and Click Edit media.
1. Add Alt or caption text.
1. Go back and you don’t see the alt or caption text.

westonruter:
*Work in progress*

This PR will introduce a lot of churn but it will give a much more extensible foundation for us to extend new media widget.

Fixes #8 
Fixes #7
Fixes #16


westonruter:
@obenland FYI: Here's the progress I've made so far on refactoring the JS to be extensible, with a base `Backbone.View` that each media type can extend. Lots more work to do, but I think this will provide us with a better foundation for this JS-driven control.

obenland:
This is quite a PR. Are you still working on it? Do you think this could be split up in smaller PRs?

wp-admin/js/widgets/media-widgets.js
```diff
@@ -0,0 +1,91 @@
...
+	// } ));
+	component.widgetControls = {};
+
+	// component.registerWidget = function registerWidget( Control ) {
+	//
+	// };
+
+	$( document ).on( 'widget-added', function( event, widgetContainer ) {
+		console.info( widgetContainer )
+		var widgetContent, controlContainer, widgetForm, widgetId, idBase, ControlConstructor, ModelConstructor, modelAttributes, control, model;
```
obenland: Is the order of variables organized or random?

westonruter:
> This is quite a PR. Are you still working on it? Do you think this could be split up in smaller PRs?

Unfortunately, not. It can't be split up because it's a rewrite of the JS.

I am still working on it, and hope to have the changes I want to make in place be EOD today. But we should probably merge it once the image widget is completed, and then open new PRs for the audio and video widgets, do you think?

wp-admin/js/widgets/media-widgets.js
westonruter: This whole function is a mess and needs further cleanup.

obenland:
Is there any way I can contribute or help? Everything else seems blocked until this gets in

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -0,0 +1,174 @@
...
+		/**
+		 * Render preview.
+		 *
+		 * @returns {void}
+		 */
+		renderPreview: function renderPreview() {
+			var control = this, previewContainer, previewTemplate;
+			previewContainer = control.$el.find( '.media-widget-preview .rendered' );
+			previewTemplate = wp.template( 'wp-media-widget-image-preview' );
+			previewContainer.html( previewTemplate( { attachment: control.selectedAttachment.attributes } ) );
```
obenland, line 22: Is there a reason `control.selectedAttachment.attributes` is not passed straight to the template?

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -27,65 +27,233 @@ public function __construct() {
...
+			sprintf(
+				'
+					wp.mediaWidgets.controlConstructors[ %1$s ].prototype.mime_type = %2$s;
+					_.extend( wp.mediaWidgets.controlConstructors[ %1$s ].prototype.l10n, %3$s );
+				',
+				wp_json_encode( $this->id_base ),
+				wp_json_encode( $this->widget_options['mime_type'] ),
+				wp_json_encode( $this->l10n )
+			)
+		);
```
obenland, line 237: What made you decide to use inline scripts rather than `wp_localize_script()` and assigning its values?

westonruter: @obenland yes, by using `wp_add_inline_script()` it means we can avoid having to create a new global variable and can instead amend an existing object directly. Also, `wp_localize_script()` will cast scalar array values to strings which may not always be desired.

wp-admin/js/widgets/media-image-widget.js
westonruter: @obenland Good point. Maybe not, although maybe this is anticipating the ability to pass in a `url` separately in the case of providing an external image by URL instead of from the media library.

wp-includes/widgets/class-wp-widget-image.php
obenland: Avoiding a global variable doesn't come free. It makes `media-image-widget.js` that much harder to grok for contributors who are unfamiliar with the code

PR #18, opened 2017-03-14
Introduce instance schema to declaratively sanitize instance props

westonruter:
Re-use REST API infrastructure for validating and sanitizing instance properties. This allows us to eliminate having to override the `update` method for media widget subclasses.

This draws inspiration from:
* [#35574](https://core.trac.wordpress.org/ticket/35574): Add REST API JSON schema information to WP_Widget
* [JS Widgets](https://github.com/xwp/wp-js-widgets): [`WP_JS_Widget::get_item_schema()`](https://github.com/xwp/wp-js-widgets/blob/0.4.1/php/class-wp-js-widget.php), though the JS Widgets plugin also has widget endpoints which can then be [used](https://github.com/xwp/wp-js-widgets/blob/0.4.1/php/class-js-widgets-plugin.php#L632-L650) to create `PUT` requests for letting `WP_REST_Request::has_valid_params()` and `WP_REST_Request::sanitize_params()` do the validation and sanitization for us. In the media widget there is no endpoint (yet) so we have to iterate over the schema ourselves to validate and sanitize the instance props.

### DIFF
--- a/wp-admin/js/widgets/media-image-widget.js
+++ b/wp-admin/js/widgets/media-image-widget.js
@@ -40,7 +40,7 @@
 
 			props = {
 				attachment_id: attachment.id,
-				url: attachment.url,
+				url: attachment.sizes[ displaySettings.size ].url,
 				size: displaySettings.size,
 				width: 0, // Reset.
 				height: 0, // Reset.
--- a/wp-admin/js/widgets/media-image-widget.js
+++ b/wp-admin/js/widgets/media-image-widget.js
@@ -48,7 +48,7 @@
 				caption: attachment.caption,
 				alt: attachment.alt,
 				link_type: displaySettings.link,
-				link_url: displaySettings.link_url
+				link_url: displaySettings.linkUrl
 			};
 
 			return props;
--- a/wp-admin/js/widgets/media-image-widget.js
+++ b/wp-admin/js/widgets/media-image-widget.js
@@ -4,7 +4,6 @@
 
 	var ImageWidgetModel, ImageWidgetControl;
 
-	// @todo Implement validate methods to force the right type.
 	// Defaults will get set via WP_Widget_Image::enqueue_admin_scripts().
 	ImageWidgetModel = component.MediaWidgetModel.extend( {} );
 
--- a/wp-includes/widgets/class-wp-widget-image.php
+++ b/wp-includes/widgets/class-wp-widget-image.php
@@ -28,16 +28,9 @@ class WP_Widget_Image extends WP_Widget_Media {
 		return array_merge(
 			parent::get_instance_schema(),
 			array(
-				'attachment_id' => array(
-					'type' => 'integer',
-					'default' => 0,
-				),
-				'url' => array(
-					'type' => 'string',
-					'default' => '',
-				),
 				'size' => array(
-					'type' => 'string', // @todo enum.
+					'type' => 'string',
+					'enum' => array_merge( get_intermediate_image_sizes(), array( 'full', 'custom' ) ),
 					'default' => 'full',
 				),
 				'width' => array( // Via 'customWidth', only when size=custom; otherwise via 'width'.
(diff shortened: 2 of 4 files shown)


wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -127,7 +137,7 @@ public function update( $new_instance, $instance ) {
 	 * @return void
 	 */
 	public function render_media( $instance ) {
-		$instance = array_merge( $this->default_instance, $instance );
+		$instance = array_merge( wp_list_pluck( $this->get_instance_schema(), 'default' ), $instance );
```
obenland, line 140: I think we could avoid the majority of `wp_list_pluck()` calls by doing

```php
// In WP_Widget_Media:
public function __construct() {
	// ...

	$this->default_instance = wp_list_pluck( $this->get_instance_schema(), 'default' );
}
```

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -17,16 +17,32 @@
...
+	 *
+	 * @link https://core.trac.wordpress.org/ticket/35574
+	 * @return array
 	 */
-	protected $default_instance = array(
-		'attachment_id' => 0,
-		'url' => '',
-		'title' => '',
-	);
+	protected function get_instance_schema() {
```
obenland: Is there a reason this is a method and not a property? Are we going to compute anything here?

westonruter: @obenland Yes, merging with the parent `get_instance_schema`. See the image widget. In that widget you can see that it's also making some calls in the schema definition.

wp-includes/widgets/class-wp-widget-image.php
westonruter: @obenland yes, your right. But since there's only a couple calls I thought maybe it would be better keep it DRY instead of duplicate the data in two places.

wp-includes/widgets/class-wp-widget-media.php
obenland: Could we move it under `__construct` then? Probably even under the public methods if we want to order them that way

wp-includes/widgets/class-wp-widget-image.php
obenland: Why does it need to be part of the schema in the first place? Neither `rest_validate_value_from_schema()` nor `rest_sanitize_value_from_schema()` use it and it adds the overhead of having to pull it

wp-includes/widgets/class-wp-widget-media.php
westonruter: Good call. I moved it below `__construct`. I want to keep it toward the top though since it is an `!important` method, since it drives so much of the behavior of the widget.

wp-includes/widgets/class-wp-widget-image.php
westonruter: Because it's part of the format for JSON Schema used in the REST API, anticipating a future move toward exposing these widgets in the REST API. Also, we avoid having to copy the keys down twice if we organize them under one  schema array instead of two separate arrays. Anyway, I think `wp_list_pluck` is fast so I'm not concerned about optimizing for it.

obenland:
Testing the branch I noticed a few bugs. When selecting a caption image in thumbnail size:
* the customizer preview uses a medium size image with a thumbnail size caption
* the edit modal also shows the full size image
* the edit modal is missing link information and CSS classes etc

westonruter:
@obenland I can't seem to replicate those issues:

I tried selecting an image from the media library:

image

Upon selection it rendered as follows with the instance data `var_dump()`'ed out:

image

Then clicking Edit Image I then saw the modal populated with:

image

And I then populated it with: 

image

And the result was:

image

obenland:
This one should show a thumbnail:

image

This one shows a medium sized image in 150px width. It's not square:

image

I fixed it in 1da9fac4b477326f0f67513f58482c1205489f5e. Albeit not to Travis' satisfaction.

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -40,7 +40,7 @@
 
 			props = {
 				attachment_id: attachment.id,
-				url: attachment.url,
+				url: attachment.sizes[ displaySettings.size ].url,
```
westonruter, line 42: Good catch.

westonruter:
@obenland the build error from Travis is due to some glitch with Git that crops up from time to time, re: Invalid symmetric reference expression. Pushing another commit seems to resolve the issue 🤷‍♂️

westonruter:
…and Travis gives us a ✅

obenland (approved):

PR #19, opened 2017-03-15
Fix widget initialization on admin screen

westonruter:
PR currently targeting merge into feature branch for still-open #18. Once merged, target branch can be changed to `master`.

* Improves JS initialization logic, removing self-booting JS and improving structure.
* Add syncing of server-sanitized widget data into Backbone model upon `widget-updated`.

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -220,9 +220,6 @@ wp.mediaWidgets = ( function( $ ) {
 
 	component.MediaWidgetModel = Backbone.Model.extend( {
 		schema: {
-			id: {
-				type: 'string'
-			},
 			title: {
 				type: 'string',
 				'default': ''
@@ -280,6 +277,7 @@ wp.mediaWidgets = ( function( $ ) {
 			_.each( attrs, function( value, name ) { // eslint-disable-line complexity
 				var type;
 				if ( ! model.schema[ name ] ) {
+					castedAttrs[ name ] = value;
 					return;
 				}
 				type = model.schema[ name ].type;
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -298,10 +298,16 @@ wp.mediaWidgets = ( function( $ ) {
 	 * @returns {void}
 	 */
 	component.handleWidgetAdded = function handleWidgetAdded( event, widgetContainer ) {
-		var widgetContent, controlContainer, widgetForm, idBase, ControlConstructor, ModelConstructor, modelAttributes, widgetControl, widgetModel;
-		widgetForm = widgetContainer.find( '> .widget-inside > .form' );
+		var widgetContent, controlContainer, widgetForm, idBase, ControlConstructor, ModelConstructor, modelAttributes, widgetControl, widgetModel, widgetId;
+		widgetForm = widgetContainer.find( '> .widget-inside > .form, > .widget-inside > form' );
 		widgetContent = widgetForm.find( '> .widget-content' );
 		idBase = widgetForm.find( '> .id_base' ).val();
+		widgetId = widgetForm.find( '> .widget-id' ).val();
+
+		// Prevent initializing already-added widgets.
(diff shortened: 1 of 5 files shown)


wp-admin/js/widgets/media-widgets.js
```diff
@@ -57,54 +60,79 @@ wp.mediaWidgets = ( function( $ ) {
...
-				control.selectedAttachment.set( {
-					id: control.model.get( 'attachment_id' )
-				} );
-				control.selectedAttachment.fetch().done( function() {
-					control.attachmentFetched.resolve();
-				} );
-			}
+			// Re-render the preview when the attachment changes.
+			control.selectedAttachment = new wp.media.model.Attachment( { id: 0 } );
+			control.listenTo( control.selectedAttachment, 'change', function() {
```
obenland: This could probably be simplified to
```js
control.listenTo( control.selectedAttachment, 'change', control.renderPreview );
```

wp-admin/js/widgets/media-widgets.js
```diff
@@ -303,38 +341,117 @@ wp.mediaWidgets = ( function( $ ) {
...
+	 * a change to a widget in the customizer.
+	 *
+	 * @param {jQuery.Event} event - Event.
+	 * @param {jQuery}       widgetContainer - Widget container element.
+	 * @returns {void}
+	 */
+	component.handleWidgetUpdated = function handleWidgetUpdated( event, widgetContainer ) {
+		var widgetForm, widgetContent, widgetId, widgetControl, attributes = {};
+		widgetForm = widgetContainer.find( '> .widget-inside > .form, > .widget-inside > form' );
+		widgetContent = widgetForm.find( '> .widget-content' );
```
obenland, line 470: Are there instance properties inside `widgetForm` that are not part of `widgetContent`? If not we could just use `widgetForm` in line 408

wp-admin/js/widgets/media-widgets.js
```diff
@@ -303,38 +341,117 @@ wp.mediaWidgets = ( function( $ ) {
...
+	 *
+	 * This function exists to prevent the JS file from having to boot itself.
+	 * When WordPress enqueues this script, it should have an inline script
+	 * attached which calls wp.mediaWidgets.init().
+	 *
+	 * @returns {void}
+	 */
+	component.init = function init() {
+		$( document ).on( 'widget-added', component.handleWidgetAdded );
+		$( document ).on( 'widget-synced widget-updated', component.handleWidgetUpdated );
```
obenland: We could string the `on()` calls together and avoid a second selector call

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -151,6 +153,9 @@ public function update( $new_instance, $instance ) {
 			if ( ! array_key_exists( $field, $new_instance ) ) {
 				continue;
 			}
+			if ( ! empty( $field_schema['readonly'] ) ) {
```
obenland: Is this used anywhere?

obenland (approved):
I noticed when changing images, that the "Open link in new tab" option value is being taken over from the previous image since it doesn't have a setting in the media modal. Not sure if that should be fixed with this PR, but we'd have to make sure that we use default settings when selecting a new image.

Looks good otherwise

wp-includes/widgets/class-wp-widget-media.php
westonruter: No it's not. I'll fix.

wp-admin/js/widgets/media-widgets.js
westonruter: All of the hidden inputs for the instance properties will be inside of `.widget-content`, as it is what `WP_Media_Widget::form()` renders. I suppose we could just use `widgetForm` but it seemed like it would be better to restrict the scope of the selector query if we know they're only ever going to be inside that element.

wp-admin/js/widgets/media-widgets.js
westonruter: Good point! Yes, if we add `renderPreview` up above to the `bindAll` call. I'll do that.

PR #20, opened 2017-03-15
Reset props when selecting a new image

obenland:
Also orders props alphabetically for easier scanning.

See #19.
/cc @westonruter

### DIFF
--- a/wp-admin/js/widgets/media-image-widget.js
+++ b/wp-admin/js/widgets/media-image-widget.js
@@ -39,15 +39,20 @@
 
 			props = {
 				attachment_id: attachment.id,
-				url: attachment.sizes[ displaySettings.size ].url,
-				size: displaySettings.size,
-				width: 0, // Reset.
-				height: 0, // Reset.
 				align: displaySettings.align,
-				caption: attachment.caption,
 				alt: attachment.alt,
+				caption: attachment.caption,
+				image_classes: '',
+				image_title: '',
+				link_classes: '',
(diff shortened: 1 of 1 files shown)


wp-admin/js/widgets/media-image-widget.js
```diff
@@ -39,15 +39,20 @@
...
 				align: displaySettings.align,
-				caption: attachment.caption,
 				alt: attachment.alt,
+				caption: attachment.caption,
+				image_classes: '',
+				image_title: '',
+				link_classes: '',
+				link_rel: '',
+				link_url: displaySettings.linkUrl,
+				link_target_blank: false,
```
westonruter, line 50: @obenland I think that any props for display settings not shown in the select frame should not get reset:

image

So I think that the `align`, `link_type`, `link_url`, and `size` (and `width`, `height`, `url`) should all get reset or overridden from `displaySettings` but everything else can persist across image selections so that the user doesn't have to re-supply them. What do you think?

obenland: I don't know, it just came completely unexpected. I associate the modal settings with the image, not the widget. The `title` for example is something that I'm sure doesn't carry over from one image to the other, much like custom css classes.

I'd expect the settings to be like they are when first setting up the widget and selecting an image

westonruter: I guess we'll have to let user testing decide! 😄

PR #21, opened 2017-03-15
Pre-populate attachment display settings in select frame with widget instance state

westonruter:
To preserve previously-supplied attachment display settings, restore them when opening the media frame to change the selected image.

Also focus on first focusable element select and edit media frames for improved accessibility.

### DIFF
--- a/wp-admin/js/widgets/media-image-widget.js
+++ b/wp-admin/js/widgets/media-image-widget.js
@@ -80,7 +80,7 @@
 		 * @returns {void}
 		 */
 		selectMedia: function selectMedia( event ) {
-			var control = this, selection, mediaFrame, library, CustomizedDisplaySettingsLibrary, customizedDisplaySettings;
+			var control = this, selection, mediaFrame, CustomizedDisplaySettingsLibrary, customizedDisplaySettings;
 
 			event.preventDefault();
 
@@ -142,24 +142,22 @@
 				}
 			} );
 
-			library = new CustomizedDisplaySettingsLibrary( {
-				library: wp.media.query( {
-					type: control.mime_type
-				} ),
-				title: control.l10n.select_media,
-				selection: selection,
-				multiple: false,
-				priority: 20,
-				display: true, // Attachment display setting.
-				filterable: false
-			} );
-
 			mediaFrame = wp.media( {
 				frame: 'select',
 				button: {
 					text: control.l10n.add_to_widget
 				},
-				states: library
+				states: new CustomizedDisplaySettingsLibrary( {
+					library: wp.media.query( {
+						type: control.mime_type
+					} ),
+					title: control.l10n.select_media,
+					selection: selection,
+					multiple: false,
+					priority: 20,
+					display: true, // Attachment display setting.
+					filterable: false
+				} )
 			} );
 
 			// Handle selection of a media item.
--- a/wp-admin/js/widgets/media-image-widget.js
+++ b/wp-admin/js/widgets/media-image-widget.js
@@ -165,11 +165,6 @@
 				url: control.model.get( 'url' )
 			};
 
-			wp.media.events.trigger( 'editor:image-edit', {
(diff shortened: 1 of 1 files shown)


wp-admin/js/widgets/media-image-widget.js
```diff
@@ -113,6 +113,30 @@
 			} );
 
 			mediaFrame.open();
+
+			// @todo There must be a better way to access this view.
+			displaySettingsView = mediaFrame.views.get( '.media-frame-content' )[0].sidebar._views.display;
```
westonruter: @adamsilverstein There is a better way to do this, right?

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -141,11 +165,6 @@
 				url: control.model.get( 'url' )
 			};
 
-			wp.media.events.trigger( 'editor:image-edit', {
-				metadata: metadata,
-				image: control.$el.find( 'img:first' )
-			} );
-
```
westonruter, line 162: I could find no purpose for this, other than being copy/pasted from the another place in core.

wp-admin/js/widgets/media-image-widget.js
westonruter: Oh, looks like `mediaFrame.views.get( '.media-frame-content' )[0].sidebar.get( 'display' )`

obenland:
I don't think that's working? They get reset when I select a new image, note the alignment:
untitled

westonruter:
@obenland right, the settings are changing when you select a new image, but at least they are populated the same when the media frame first opens, right? In this way a user would be able to adjust the settings they previously set from this frame in the same UI.

It seems the missing piece is to have the settings not be overridden when selecting a different image?

obenland:
> It seems the missing piece is to have the settings not be overridden when selecting a different image?

If you want to preserve settings, yes

westonruter:
So then this is closely related to #20.

@melchoyce thoughts on the persistence of the selected alignment, link, and image size when you chose another image? Should they get reset? Should the previously-selected display settings be re-populated in the select frame when you click Change Image?

melchoyce:
To clarify:
1. Select an image
2. Change its alignment, where it links to, or size
3. Add image to widget 
4. Change image

You're asking:
1. Should the image you selected stay selected, OR clear so no image is selected?
2. Should its alignment, link, and size stay the same, OR reset to defaults?

obenland:
The latter, should attributes carry over or be reset to defaults

melchoyce:
I think yes, attributes should carry over. The only exception I can think of is the link attribute — if you've selected "link to this specific URL," you might not want that particular URL to remain if you change the image. (Happy to hear alternate points of view on this.)

@karmatosed do you have any thoughts here?

karmatosed:
I would second that attributes should carry over. I can see a case for linking URLs not being, but feel it makes more sense to carry all over, including those.

westonruter:
This is ready for review!

westonruter:
The link URL will only carry-over when a `custom` type is selected. If the user selected to link to the file or to the attachment page, then the default URL will be used instead.

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -80,30 +80,89 @@
...
+					display.set( customizedDisplaySettings.attributes );
+					if ( 'custom' === customizedDisplaySettings.get( 'link_type' ) ) {
+						display.linkUrl = customizedDisplaySettings.get( 'link_url' );
+					}
+					display.on( 'change', this.handleDisplaySettingChange );
+					return display;
+				}
+			} );
+
+			library = new CustomizedDisplaySettingsLibrary( {
```
obenland: Any objections against bringing this back inline?

It would make a 23 line change a two line change (and it would save  us a variable)

obenland (approved):
Works well, good job!

Issue #15, opened 2017-03-11
Button text can go over the button


samikeijonen:
In different languages words can be longer. In Finnish button text go over the button.

<img width="263" alt="text on button" src="https://cloud.githubusercontent.com/assets/1820415/23824785/a10f759c-0685-11e7-8616-9a622b546105.png">

westonruter:
I used flexbox to try to make the two buttons equal width. I'm not a flexbox expert, but we could use `flex-grow` to allow the buttons to fully take up the width, but resize themselves when one of the buttons has more text. So in this example the left button maybe could be wider than the right button. Otherwise, an ellipsis could be used. Or in such languages that have longer strings the translation may need to just be “Change” as opposed to “Change media”.

PR #23, opened 2017-03-16
Update Buttons Flexbox

timmyc:
After looking at #15 - I decided to play with the flexbox setup on the buttons a bit.  This small change just keeps the buttons naturally sized in-lieu of growing to fit the entire space on the widget screen, and adds `flex-wrap` to handle any situations where long translations would cause the buttons to get too large for one line.

__Before__
before

select-before

__After__
edit-after

select-both

### DIFF
--- a/wp-admin/css/widgets/media-widgets.css
+++ b/wp-admin/css/widgets/media-widgets.css
@@ -10,15 +10,16 @@
 	display: -ms-flexbox;
 	display: -webkit-flex;
 	display: flex;
+	flex-wrap: wrap;
 }
 
 .media-widget-buttons .button {
-	-webkit-box-flex: 1;
-	-webkit-flex-grow: 1;
-	-moz-box-flex: 1;
-	-ms-flex-positive: 1;
-	-ms-flex: 1;
-	flex-grow: 1;
+	-webkit-box-flex: 0 0 auto;
+	-webkit-flex-grow: 0 0 auto;
+	-moz-box-flex: 0 0 auto;
+	-ms-flex-positive: 0 0 auto;
+	-ms-flex: 0 0 auto;
+	flex-grow: 0 0 auto;
 }
 .media-widget-buttons .button {
 	margin-left: 10px;


westonruter:
In #24 I've suggested the image should be centered when it isn't the full width.

I think it is preferred if the buttons stretch the full width or 50% if there are two. Compare with the media control in the customizer:

image

timmyc:
Gotcha, makes sense.  The Select Image button still feels pretty large to me - but perhaps a designer can chime in on the matter ;)  - I'll close this one out for now.

westonruter:
You know what? You're right. In the customizer, when only one button is displayed it gets shown at the same width and floated to the right:

image

@melchoyce should this be replicated here?

The following has been reported in #29.

----

I can also see that we may be missing styling for the non-selected state, where we don't currently show that dashed border, though this seems to imply it is a drag-and-droppable region (which it isn't in core, yet per [#35827](https://core.trac.wordpress.org/ticket/35827)):

image

timmyc:
From Slack convo, here are the widget states with right-aligned buttons /cc @melchoyce 

melchoyce:
Okay, yeah, definitely has to be left-aligned. Thanks @timmyc.

westonruter:
@timmyc note that the image would be center-aligned. Also, there is now a placeholder dotted box now. I just updated from `master` so you should see that now. This will help with the buttons feeling orphaned on the right.

And what about designs for letting the buttons be 100% wide (when one button), and 50% wide if two? Or 50% wide always, with the Edit Image button hidden when no selection is made and the Select Image button floated to the right?

timmyc:
@westonruter I'll rebase my branch and add some screen grabs for the other approaches too.

timmyc:
Rebased - new screenshots aligned right, customizer:

Widget admin looks same as above - perhaps I'll add the dotted placeholder there.

westonruter:
I still think having the buttons 50% wide feels more refined (like intentionally designed), and is inline with the customizer media control. When the buttons have different widths to me it feels not as sharp.

timmyc:
Customizer, current for comparison with the intial design.

melchoyce:
@westonruter What would you do in cases where the button label translation is longer than the width of the button?

westonruter:
@melchoyce in that case it's the same situation as with the existing media control. Translators have to know the context and know the size of the string they choose. This could be indicated via the `translators` comment or via the `$context` parameter in the translation function. This is what we did for the “Hide Controls” label in [#38762](https://core.trac.wordpress.org/ticket/38762). See [r39214](https://core.trac.wordpress.org/changeset/39214).

melchoyce:
Let's go with the 50%/100% width buttons, then, and consider re-examining this pattern in core.

westonruter:
See #38.

PR #26, opened 2017-03-16
Ensure preview image in control is center-aligned and does not overflow

### LINKED ISSUE #24: Image doesn't sit within widget container
Just tried this new Image Widget on a custom theme, Twenty Seventeen and also on Twenty Twelve and got the same result after adding an image. The image doesn't sit nicely in the actual widget.

Here's Twenty Twelve - https://cl.ly/3z0n3x0j0n15

westonruter:
Fixes #24.

Centers image that is smaller than the control width:

screen shot 2017-03-15 at 21 53 42

And adds `max-width` to prevent an image that is wider than the control from overflowing:

screen shot 2017-03-15 at 21 59 39

### DIFF
--- a/wp-admin/css/widgets/media-widgets.css
+++ b/wp-admin/css/widgets/media-widgets.css
@@ -30,3 +30,6 @@
 .media-widget-control .media-widget-preview {
 	text-align: center;
 }
+.media-widget-control .media-widget-preview img {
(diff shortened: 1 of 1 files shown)


obenland (approved):

obenland:
Thanks for the screenshots!

PR #27, opened 2017-03-16
Handle deletion of attachment associated with media widget

### LINKED ISSUE #22: Deleted Image from Media Library results in Broken Widget
Not certain if we should be proactive about this flow or not but the following results in a broken image widget:

__Steps to Recreate__
- Create a new Image Widget instance.
- Add a photo from the library, or upload a new item
- Save, verify all looks happy
- Open the media library, and permanently delete the image used in the widget

The Edit Image button still is shown in the widget editor, and selecting it results in:

On the front-end, it appears only the image title is rendered, which is good, but should we not render anything when there is no image to show?

wordpress_develop_ _just_another_wordpress_site

obenland:
Shows an error message in the media widget when its associated attachment is deleted.

Also adds a label to the attachment in the media list, when its used in a widget.

Fixes #22.

/cc @timmyc

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -73,7 +73,7 @@ wp.mediaWidgets = ( function( $ ) {
 		 * @type {Object}
 		 */
 		events: {
-			'click .notice a': 'selectMedia',
+			'click .notice-missing-attachment a': 'selectMedia',
 			'click .select-media': 'selectMedia',
 			'click .edit-media': 'editMedia'
 		},
@@ -155,10 +155,10 @@ wp.mediaWidgets = ( function( $ ) {
 				} );
 				attachment.fetch()
 					.done( function() {
-						control.selectedAttachment.set( attachment.attributes );
+						control.selectedAttachment.set( _.extend( {}, attachment.attributes, { error: false } ) );
 					} )
 					.fail( function() {
-						control.selectedAttachment.set( { error: true } );
+						control.selectedAttachment.set( { error: 'missing_attachment' } );
 					} );
 			}
 		},
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -152,9 +152,13 @@ wp.mediaWidgets = ( function( $ ) {
 				attachment = new wp.media.model.Attachment( {
 					id: control.model.get( 'attachment_id' )
 				} );
-				attachment.fetch().done( function() {
-					control.selectedAttachment.set( attachment.attributes );
-				} );
+				attachment.fetch()
+					.done( function() {
+						control.selectedAttachment.set( attachment.attributes );
+					} )
+					.fail( function() {
+						control.selectedAttachment.set( { error: true } );
+					} );
 			}
 		},
 
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -73,6 +73,7 @@ wp.mediaWidgets = ( function( $ ) {
 		 * @type {Object}
 		 */
 		events: {
+			'click .notice a': 'selectMedia',
 			'click .select-media': 'selectMedia',
 			'click .edit-media': 'editMedia'
 		},
--- a/wp-includes/widgets/class-wp-widget-image.php
+++ b/wp-includes/widgets/class-wp-widget-image.php
@@ -36,8 +36,8 @@ class WP_Widget_Image extends WP_Widget_Media {
 				__( 'We can&#8217;t find that image. Check your <a href="%s">media library</a> and make sure it wasn&#8217;t deleted.' ),
 				esc_url( admin_url( 'upload.php' ) )
 			),
-			/* translators: %s is widget count */
-			'media_library_state' => _n_noop( 'Image Widget', 'Image Widget (%s)' ),
+			/* translators: %d is widget count */
+			'media_library_state' => _n_noop( 'Image Widget (%d instance)', 'Image Widget (%d instances)' ),
 			'no_media_selected' => __( 'No image selected' ),
 			'select_media' => __( 'Select Image' ),
 		) );
--- a/wp-includes/widgets/class-wp-widget-image.php
+++ b/wp-includes/widgets/class-wp-widget-image.php
@@ -34,8 +34,6 @@ class WP_Widget_Image extends WP_Widget_Media {
 			'change_media' => __( 'Change Image' ),
 			'select_media' => __( 'Select Image' ),
 		) );
-
-		add_filter( 'display_media_states', array( $this, 'display_media_state' ), 10, 2 );
 	}
 
 	/**
@@ -212,41 +210,6 @@ class WP_Widget_Image extends WP_Widget_Media {
 		}
 
 		echo $image;
-	}
-
-	/**
-	 * Filters the default media display states for items in the Media list table.
-	 *
-	 * @since 4.8.0
-	 * @access public
-	 *
-	 * @param array   $states An array of media states.
-	 * @param WP_Post $post   The current attachment object.
-	 * @return array
-	 */
-	public function display_media_state( $states, $post ) {
-
-		// Short-circuit if the post is not an image attachment.
-		if ( 'attachment' !== $post->post_type || 0 !== strpos( $post->post_mime_type, 'image/' ) ) {
-			return $states;
-		}
-
-		// Count how many times this image attachment is used in image widgets.
-		$image_used_count = 0;
-		foreach ( $this->get_settings() as $instance ) {
-			if ( isset( $instance['attachment_id'] ) && $instance['attachment_id'] === $post->ID ) {
-				$image_used_count++;
-			}
-		}
-
-		if ( $image_used_count > 1 ) {
-			/* translators: %d is widget count */
-			$states[] = sprintf( __( 'Image Widgets (%d)' ), $image_used_count );
-		} elseif ( 1 === $image_used_count ) {
-			$states[] = __( 'Image Widget' );
-		}
-
-		return $states;
 	}
 
 	/**
--- a/wp-includes/widgets/class-wp-widget-image.php
+++ b/wp-includes/widgets/class-wp-widget-image.php
@@ -33,7 +33,10 @@ class WP_Widget_Image extends WP_Widget_Media {
 			'edit_media' => __( 'Edit Image' ),
 			'change_media' => __( 'Change Image' ),
 			'select_media' => __( 'Select Image' ),
-			'error' => __( 'Image could not be loaded. Please verify it was not deleted.' ),
+			'error' => sprintf(
+				__( 'We can&#8217;t find that image. Check your <a href="%s">media library</a> and make sure it wasn&#8217;t deleted.' ),
+				esc_url( admin_url( 'upload.php' ) )
+			),
 		) );
 	}
 
@@ -269,7 +272,7 @@ class WP_Widget_Image extends WP_Widget_Media {
 				<img class="attachment-thumb" src="{{ data.attachment.sizes.full.url }}" draggable="false" alt="" />
 			<# } else if ( data.attachment.error ) { #>
 				<div class="notice notice-error notice-alt">
-					<p><?php echo esc_html( $this->l10n['error'] ); ?></p>
+					<p><?php echo $this->l10n['error']; ?></p>
 				</div>
 			<# } #>
 		</script>
--- a/wp-includes/widgets/class-wp-widget-image.php
+++ b/wp-includes/widgets/class-wp-widget-image.php
@@ -33,6 +33,7 @@ class WP_Widget_Image extends WP_Widget_Media {
 			'edit_media' => __( 'Edit Image' ),
 			'change_media' => __( 'Change Image' ),
 			'select_media' => __( 'Select Image' ),
+			'error' => __( 'Image could not be loaded. Please verify it was not deleted.' ),
 		) );
 	}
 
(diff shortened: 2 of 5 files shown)


wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -213,6 +215,27 @@ public function render_media( $instance ) {
...
+	 * @since 4.8.0
+	 * @access public
+	 *
+	 * @param array   $states An array of media states.
+	 * @param WP_Post $post   The current attachment object.
+	 * @return array
+	 */
+	public function display_media_state( $states, $post ) {
+		$settings      = $this->get_settings();
+		$attachment_id = empty( $settings[ $this->number ]['attachment_id'] ) ? 0 : $settings[ $this->number ]['attachment_id'];
```
westonruter: This is only going to work if you have one image widget. If you have multiple image widgets, then you'd have to iterate over all `$settings` to check each one to see if it has `attachment_id` that is the same as `$post->ID`. As it is right now, it's only looking at the current `$this->number` widget instance, which is probably going to be the last one registered.

westonruter: Fixed in db981150.

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -213,6 +215,41 @@ public function render_media( $instance ) {
...
+	 * Filters the default media display states for items in the Media list table.
+	 *
+	 * @since 4.8.0
+	 * @access public
+	 *
+	 * @param array   $states An array of media states.
+	 * @param WP_Post $post   The current attachment object.
+	 * @return array
+	 */
+	public function display_media_state( $states, $post ) {
```
westonruter: Should this maybe be moved to the base `WP_Widget_Media` class? Naturally some of the hard-coded values here would need to change, like replace `'image/'` with `$this->widget_options['mime_type']` and use `$this->name` instead of `'Image Widget'`.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -234,6 +235,45 @@ class="media-widget-instance-property"
...
+	 * Deletes the Widget when the attachment file is deleted.
+	 *
+	 * @since 4.8.0
+	 * @access public
+	 *
+	 * @param int $post_id Attachment ID.
+	 */
+	public function delete_attachment_data( $post_id ) {
+		$settings      = $this->get_settings();
+		$attachment_id = empty( $settings[ $this->number ]['attachment_id'] ) ? 0 : $settings[ $this->number ]['attachment_id'];
```
westonruter: As done in `display_media_state `, this will need to loop over all `$settings` not just `$settings[ $this->number ]`.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -234,6 +235,45 @@ class="media-widget-instance-property"
...
 
 	/**
+	 * Deletes the Widget when the attachment file is deleted.
+	 *
+	 * @since 4.8.0
+	 * @access public
+	 *
+	 * @param int $post_id Attachment ID.
+	 */
+	public function delete_attachment_data( $post_id ) {
```
westonruter: Deleting the widget instances when the associated attachment is deleted makes me nervous. I think it would be better to instead cause the widget to not render anything, and show a notification in the widget control that the attachment has been removed. For example, I think it should behave like an RSS widget pointing to a 404 feed URL.

By showing a notice that the attachment has been deleted, there would be less user mystery when they find a widget disappeared. Deleting the widget would also cause the widget's title to be dropped, resulting in data loss. I think it would be better if a user discovers they deleted an image used by an image widget, and then picks another image for that existing widget, rather than having to create a whole new widget from scratch.

westonruter (changes requested):
Something else related to the behavior of a widget upon deletion of its associated attachment is what happens when you try deleting the image used for the image widget you're currently editing. See https://cloudup.com/cJs4gRGYnXE

Per https://github.com/xwp/wp-core-media-widgets/pull/27/files#r106510104 I think this should result in the selected image being removed from the widget, and a notification shown that the image was removed.

wp-includes/widgets/class-wp-widget-media.php
timmyc: ➕  I was thinking the same thing - when the media item is deleted, perhaps just not render anything - or perhaps if the user has a certain role, and is viewing the site, then show a notice in the widget area that the attachment is missing.

wp-includes/widgets/class-wp-widget-image.php
obenland: > use $this->name instead of 'Image Widget'

Duh! I didn't think of that. Yes it should be in `WP_Widget_Media`, the only reason it's not is because I didn't want a hard coded array of names. So yeah, that solves it.

obenland:
7d566e9 introduced an error message when an attachment can't be loaded:

screen shot 2017-03-17 at 8 57 17 am

@melchoyce What should the message look like and say?
@westonruter @timmyc Is this how you would go about it? Setting an error on the model? Also: What other instances can you think of where the attachment can't be displayed?

EDIT: We should probably not display the Edit button in case of an error either.

melchoyce:
How about:



> We can't find that image. Check your media library and make sure it wasn't deleted.

We'll also want to make sure the text is left-aligned.

(Props @michelleweber)

obenland:
Thank you! Updated in c5bc6d5

obenland:
@westonruter: Using the widget name doesn't work too well:

screen shot 2017-03-17 at 10 49 01 am

I'm not sure how I feel about `sprintf( __( '%s Widget' ), $this->name )`, even with a translator comment. Nesting translated strings like that usually doesn't go over well.

westonruter:
@obenland oh I forgot that “widget” wasn't actually part of the name. Two options I can think of:

1. We change the translation string to be more robust for translation purposes, like `__( 'Media widgets for "%s"' )`.
2. We introduce new `l10n` label for this purpose, as we already have for defining the strings used for the buttons. So then we'd use something like `$this->l10n['media_library_display_state']` in this case.

I think 2 would be a better choice.

obenland:
> I think 2 would be a better choice.

Agreed

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -33,6 +33,10 @@ public function __construct() {
 			'edit_media' => __( 'Edit Image' ),
 			'change_media' => __( 'Change Image' ),
 			'select_media' => __( 'Select Image' ),
+			'error' => sprintf(
+				__( 'We can&#8217;t find that image. Check your <a href="%s">media library</a> and make sure it wasn&#8217;t deleted.' ),
```
westonruter: The Backbone view for should have a new event handler for clicking on this link to have the same result as clicking the "Select Media" or "Change Media" buttons. The underlying `href` can remain there so they can right-click to open the media library in a new tab, but otherwise we should keep them inside the customizer and open the select frame for the media modal.

obenland: Added in 43e5c93.
I decided to go with a new selector that is a little broader than I'd like it to be, in order to not give translators the chance to break functionality by accidentally omitting the `.select-media` class.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -235,6 +244,36 @@ class="media-widget-instance-property"
...
+		$use_count = 0;
+		foreach ( $this->get_settings() as $instance ) {
+			if ( isset( $instance['attachment_id'] ) && $instance['attachment_id'] === $post->ID ) {
+				$use_count++;
+			}
+		}
+
+		if ( $use_count > 1 ) {
+			/* translators: %d is widget count */
+			$states[] = sprintf( __( 'Media Widgets (%d)' ), $use_count );
```
westonruter: Should this be changed to reference `$this->l10n['media_library_state']`?

westonruter: Should this be changed to reference `$this->l10n['media_library_state']`?

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -53,11 +55,16 @@ public function __construct( $id_base, $name, $widget_options = array(), $contro
 		$control_opts = wp_parse_args( $control_options, array() );
 
 		$l10n_defaults = array(
-			'no_media_selected' => __( 'No media selected' ),
-			'edit_media' => __( 'Edit Media' ),
+			'add_to_widget' => __( 'Add to Widget' ),
 			'change_media' => __( 'Change Media' ),
+			'edit_media' => __( 'Edit Media' ),
+			'error' => sprintf(
```
westonruter: There could be additional errors in the future, so I think this should be `attachment_missing_error` instead.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -73,6 +73,7 @@ wp.mediaWidgets = ( function( $ ) {
 		 * @type {Object}
 		 */
 		events: {
+			'click .notice a': 'selectMedia',
```
westonruter: I think the link should have a class name on it. If a subclass introduces a different error with a link in it, we'd want to make sure this event handler only handles clicks on the media library link. So either the container `. notice` element have a class like `missing-attachment-error`, or the link itself can have a `media-library` class that is then used in the selector here.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -152,9 +153,13 @@ wp.mediaWidgets = ( function( $ ) {
 				attachment = new wp.media.model.Attachment( {
 					id: control.model.get( 'attachment_id' )
 				} );
-				attachment.fetch().done( function() {
-					control.selectedAttachment.set( attachment.attributes );
-				} );
+				attachment.fetch()
+					.done( function() {
+						control.selectedAttachment.set( attachment.attributes );
```
westonruter: Should this also ensure that the `error` flag is cleared for good measure?

```js
control.selectedAttachment.set( _.extend( {}, attachment.attributes, { error: false } ) );
```

wp-admin/js/widgets/media-widgets.js
```diff
@@ -152,9 +153,13 @@ wp.mediaWidgets = ( function( $ ) {
...
+					} )
+					.fail( function() {
+						control.selectedAttachment.set( { error: true } );
```
westonruter: Maybe the `error` here should be a code like `missing_attachment` instead of a boolean to anticipate additional error types in the future.

westonruter (changes requested):

wp-admin/js/widgets/media-widgets.js
obenland: I think I prefer the former, [I'd like to avoid a situation](https://github.com/xwp/wp-core-media-widgets/pull/27#discussion_r106741038) where translators accidentally break functionality.

wp-includes/widgets/class-wp-widget-media.php
obenland: You'd want a singular and a plural version? Or just the singular?

westonruter: The plural conditional was an enhancement I came up with. Maybe it wouldn't be useful in practice, but I can imagine that it would be helpful for users to know how many widgets are using the image.

If we have both, then perhaps we'd need `media_library_state_singular_widget` and `media_library_state_multiple_widgets`

obenland:
Should have all your change requests addressed?

westonruter:
@obenland I saved an image widget that had an assigned attachment. I then deleted the underlying attachment and reloaded the customizer. When I did, I saw:

image

I would have expected there to be an error message. It seems the order of the conditionals wasn't right? I changed them in 7f32d28 and ac1239a, and now I see as expected:

image

westonruter:
Something to note is that the `get-attachment` admin-ajax request returns with just `{"success":false}` on a failure. When the request is resolved as `rejected` in Backbone, we don't know if the request failed due to the media having been deleted or due to there being a server error or even a connection failure (e.g. you just went into a tunnel).

westonruter (approved):

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -29,9 +29,16 @@ public function __construct() {
...
-			'edit_media' => __( 'Edit Image' ),
 			'change_media' => __( 'Change Image' ),
+			'edit_media' => __( 'Edit Image' ),
+			'missing_attachment' => sprintf(
+				/* translators: placeholder is URL to media library */
+				__( 'We can&#8217;t find that image. Check your <a href="%s">media library</a> and make sure it wasn&#8217;t deleted.' ),
+				esc_url( admin_url( 'upload.php' ) )
+			),
+			/* translators: %s is widget count */
+			'media_library_state' => _n_noop( 'Image Widget', 'Image Widget (%s)' ),
```
westonruter: @obenland PHP_CodeSniffer is complaining here:

> Missing singular placeholder, needed for some languages. See https://codex.wordpress.org/I18n_for_WordPress_Developers#Plurals (WordPress.WP.I18n.MissingSingularPlaceholder)

Is this valid?

westonruter: Maybe these should actually be two separate strings. 

For more about this, see https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/567#issuecomment-224788685

obenland: I don't think that's a valid concern in our case. We're not saying `One  Widget`, `2  Widgets`, the label's singular/plural version is of no concern here really.

IMO we should rather skip the rule for that line than to introduce a language workaround (which in German at least reads quite awkwardly), what do you think?

westonruter: See #39.

PR #37, opened 2017-03-20
Fix placeholder styling on widgets admin screen

### LINKED ISSUE #29: "No image selected" needs styling
This: 

Should look like this:

Looks like it's a CSS specificity issue. The placeholder has the `.placeholder` class applied, but the placeholder style is actually more specific than that:

```
.customize-control .attachment-media-view .placeholder,
.customize-control-header .placeholder
```

obenland:
Fixes #29

### DIFF
--- a/wp-admin/css/widgets/media-widgets.css
+++ b/wp-admin/css/widgets/media-widgets.css
@@ -27,6 +27,20 @@
 .media-widget-buttons .button:first-child {
 	margin-left: 0;
 }
+
+.media-widget-control .placeholder {
+	border: 1px dashed #b4b9be;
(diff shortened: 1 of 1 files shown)


obenland:
@westonruter Is `media-widgets.css` supposed to be folded into `wp-admin/css/widgets.css` when we merge?

westonruter:
@obenland that's a good question. Historically the answer would probably have always been yes. But I think we're trending now toward not making making these impossibly-large source files even larger unless we have to, and keep files separated out where it makes sense.

westonruter (approved):

PR #39, opened 2017-03-21
Restore media_library_state singular string with PHPCS rule exclusions

westonruter:
See https://github.com/xwp/wp-core-media-widgets/pull/27#pullrequestreview-27993351


obenland (approved):
❤️

westonruter:
@obenland what if we instead actually just had the one:

```php
'media_library_state' => __( 'Image Widget' ),
```

And then had a separate string that had the count?

```php
'media_library_state_count' => __( '(%d)' ),
```

And then the state would be something like:

```php
$state = $this->l10n['media_library_state'];
if ( $count > 1 ) {
    $state .= ' ' . sprintf( $this->l10n['media_library_state_count'], $count );
}
```

This would be somewhat in line with what core does now for showing comment counts in the admin. And we could avoid the PHPCS rule exclusions.

obenland:
We'd have to account for RTL as well and the context  would be lost. I like the way this PR does it

westonruter:
I don't think that RTL would be a problem because since the `(%d)` is logically after the preceding text string, it will by nature of RTL be displayed visually on the left side of the translated string.

So in English, “Image Widget (2)” is rendered in Arabic as:

image

Even though “صورة القطعة” logically comes before “(2)” in terms of the Unicode character sequence.

obenland:
I just feel like the way this PR does it feels much more elegant and cleaner

westonruter:
According to @deBhal however, it could be a problem (https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/567#issuecomment-224788685):

> I've added another new sniff for the singular problem described in https://codex.wordpress.org/I18n_for_WordPress_Developers#Plurals
_n( 'This is a subtle problem', 'These are subtle problems', $number ).
"Singular" means exactly one in English, but that's not the case in several languages.

I don't fully understand the linguistic scenarios that could be a problem here, and it seems like our usage is unique in that it isn't really a translatable sentence so that's why can exempt it. But in that case, does it really need to be plural-translated like this in the first place? 

@ocean90 Thoughts on this i18n scenario?

ocean90:
I suggest to split this into two strings, similar to https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentyseventeen/comments.php#L34-L51.

westonruter:
@obenland trivial unit test failures in `Test_WP_Widget_Image::test_constructor()`

obenland:
Looking

Issue #32, opened 2017-03-17
Implement video widget


westonruter:
The `video` widget was disabled with the refactor in #17, and the [`WP_Widget_Video`](https://github.com/xwp/wp-core-media-widgets/blob/master/wp-includes/widgets/class-wp-widget-video.php) class needs to be refactored to match what was done for the [`WP_Widget_Image`](https://github.com/xwp/wp-core-media-widgets/blob/master/wp-includes/widgets/class-wp-widget-image.php) class, including the creation of a Backbone View `VideoWidgetControl` and `VideoWidgetModel` according to [what was done for the image widget](https://github.com/xwp/wp-core-media-widgets/blob/master/wp-admin/js/widgets/media-image-widget.js).

See #31 for the same for audio.

timmyc:
I was going to dive in on this one... do we have any designs for how the outputs of `render_media` for the front-end display and the `render_preview` should look?

melchoyce:
It should use core's default video player design:

![image](https://cloud.githubusercontent.com/assets/2846578/24766091/06ee2a60-1ac8-11e7-9de8-6c67b5c9a480.png)

![image](https://cloud.githubusercontent.com/assets/2846578/24766112/1e23288e-1ac8-11e7-9d06-f8bd1c76850d.png)

westonruter:
Take note of this comment here for some complications on custom styling of the video (and audio) elements: https://core.trac.wordpress.org/ticket/32417#comment:116

melchoyce:
PS if any of y'all wanted to look at https://core.trac.wordpress.org/ticket/31139 too... ;) <3

Issue #31, opened 2017-03-17
Implement audio widget


westonruter:
The `audio` widget was disabled with the refactor in #17, and the [`WP_Widget_Audio`](https://github.com/xwp/wp-core-media-widgets/blob/master/wp-includes/widgets/class-wp-widget-audio.php) class needs to be refactored to match what was done for the [`WP_Widget_Image`](https://github.com/xwp/wp-core-media-widgets/blob/master/wp-includes/widgets/class-wp-widget-image.php) class, including the creation of a Backbone View `AudioWidgetControl` and `AudioWidgetModel` according to [what was done for the image widget](https://github.com/xwp/wp-core-media-widgets/blob/master/wp-admin/js/widgets/media-image-widget.js).

See #32 for the same for video.

melchoyce:
This is lower priority than video (#32), gallery (#62), and slideshow (#63).

PR #40, opened 2017-03-21
Move media-agnostic selectMedia method to from image widget control to base media control class

westonruter:
The method does not need to be abstract since, currently, it is not media-specific. This eliminates a widget from having to implement an abstract method for `selectMedia`. The `getSelectFrameProps` method is already supposed to be the way for a media widget subclass control to do the special handling of the media-specific props.


westonruter:
Once this is merged I think it would be good for the audio (#31) and video (#32) widgets to get revisited so we can confirm the abstraction architecture.

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -74,125 +74,6 @@
...
 		/**
-		 * Open the media select frame to chose an item.
-		 *
-		 * @param {jQuery.Event} event - Event.
-		 * @returns {void}
-		 */
-		selectMedia: function selectMedia( event ) {
-			var control = this, selection, mediaFrame, CustomizedDisplaySettingsLibrary, customizedDisplaySettings;
-
-			event.preventDefault();
```
obenland, line 85: We'll continue to need this so clicking the media library link in the error message doesn't actually take you to the media library

obenland (approved):
Looks good with the exception of the `preventDefault()` bug

PR #41, opened 2017-03-21
Add PHP unit tests

### LINKED ISSUE #35: Add PHPUnit tests
PHP unit tests (via PHPUnit) need to ensure the instance props get sanitized as expected when calling `WP_Widget_Media::update()` and also that they get rendered as expected when calling `WP_Widget_Media::widget()`.

----

Since these widgets heavily use JavaScript we should be including QUnit tests as well. Beyond PHPUnit and QUnit, I also want to take this as an opportunity to implement acceptance tests via Codecetion or Behat. For more, see [#34693](https://core.trac.wordpress.org/ticket/34693).

obenland:
Fixes #35


tests/phpunit/test-class-wp-widget-image.php
```diff
@@ -0,0 +1,230 @@
...
+
+class Test_WP_Widget_Image extends WP_UnitTestCase {
+
+	/**
+	 * @covers WP_Widget_Image::get_instance_schema
+	 */
+	function test_get_instance_schema() {
+		$WP_Widget_Image     = new ReflectionClass( 'WP_Widget_Image' );
+		$get_instance_schema = $WP_Widget_Image->getMethod( 'get_instance_schema' );
+		$get_instance_schema->setAccessible( true );
```
westonruter: Woah. You just made a protected method public!? I didn't know about this. http://php.net/manual/en/reflectionproperty.setaccessible.php

Looks like it isn't supported in PHP 5.2, so this will need to short-circuit with `$this->markTestSkipped()` if that is the PHP version.

tests/phpunit/test-class-wp-widget-image.php
```diff
@@ -0,0 +1,230 @@
...
+		$this->assertSame( $result, array( 'link_classes' => '">' ) );
+
+		// Should return valid rel text.
+		$expected = array( 'link_rel' => 'previous' );
+		$result = $widget->update( $expected, $instance );
+		$this->assertSame( $result, $expected );
+
+		// Should filter invalid rel text.
+		$result = $widget->update( array( 'link_rel' => '"><i onload="alert(\'hello\')" />' ), $instance );
+		$this->assertSame( $result, array( 'link_rel' => '">' ) );
```
westonruter: This looks like somewhat of a suspect sanitization method.

obenland: I don't know, I couldn't think of a better way to test `sanitize_text_field` off hand. Any idea for a better one?

westonruter: Maybe:

```php
$link_rel = join( ' ', array_map( 'sanitize_html_class', preg_split( '/\s+/', trim( $link_rel ) ) ) );
```

obenland: You mean as a sanitization in `update()`?

westonruter: Like 046cb014f9258d5e2ad604896b16d0e791f98d2d

obenland: Damn unit tests. Already improving our code.

obenland:
screen shot 2017-03-22 at 3 56 04 pm

I think we're done here.

tests/phpunit/test-class-wp-widget-media.php
```diff
@@ -0,0 +1,354 @@
...
+			get_post( $attachment_id ),
+			'file',
+		) );
+		$this->assertSame( '<a href="' . esc_url( wp_get_attachment_url( $attachment_id ) ) . '">Attachment Title</a>',$result );
+
+		$result = $create_link_for->invokeArgs( $this->get_mocked_class_instance(), array(
+			get_post( $attachment_id ),
+			'post',
+		) );
+		$this->assertSame( '<a href="http://example.org/?attachment_id=4">Attachment Title</a>',$result );
```
westonruter: This line fails for me locally:

```
1) Test_WP_Widget_Media::test_create_link_for
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-<a href="http://example.org/?attachment_id=4">Attachment Title</a>
+<a href="http://tests.wordpress-develop.dev/?attachment_id=4">Attachment Title</a>
```

One thing is we can't assume the site url is `http://example.org/`.  It can be specified [`wp-tests-config.php`](https://github.com/WordPress/wordpress-develop/blob/master/wp-tests-config-sample.php) to be something else by changing the `WP_TESTS_DOMAIN` constant, which is what I have done on my environment.

Secondly, the `?attachment_id=4` part here makes me nervous because it assumes there has been a fixed number of posts inserted before this point.

So should it not rather be:

```php
$this->assertSame( '<a href="' . esc_url( get_permalink( $attachment_id ) ) . '">Attachment Title</a>', $result );
```

tests/phpunit/test-class-wp-widget-image.php
```diff
@@ -0,0 +1,468 @@
...
+	function test_get_instance_schema() {
+		if ( version_compare( PHP_VERSION, '5.3', '<' ) ) {
+			return $this->markTestSkipped( 'ReflectionMethod::setAccessible is only available for PHP 5.3+' );
+		}
+
+		$wp_widget_image     = new ReflectionClass( 'WP_Widget_Image' );
+		$get_instance_schema = $wp_widget_image->getMethod( 'get_instance_schema' );
+		$get_instance_schema->setAccessible( true );
+
+		$schema = $get_instance_schema->invoke( new WP_Widget_Image );
```
westonruter: Conventions are to include parens when invoking classes, so I think this should be `new WP_Widget_Image()`.

This is something we should improve in WPCS.

tests/phpunit/test-class-wp-widget-media.php
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	/**
+	 * Test display_media_state method.
+	 *
+	 * @covers WP_Widget_Media::display_media_state()
+	 */
+	function test_display_media_state() {
+		$widget        = $this->get_mocked_class_instance();
+		$attachment_id = self::factory()->attachment->create_object( DIR_TESTDATA . '/images/canola.jpg', 0, array(
```
westonruter: Passing a string as the first argument is deprecated. The function signature for `create_object` is:

```
/**
 * Create an attachment fixture.
 *
 * @param array $args {
 *     Array of arguments. Accepts all arguments that can be passed to
 *     wp_insert_attachment(), in addition to the following:
 *     @type int    $post_parent ID of the post to which the attachment belongs.
 *     @type string $file        Path of the attached file.
 * }
 * @param int   $legacy_parent Deprecated.
 * @param array $legacy_args   Deprecated
 */
```

tests/phpunit/test-class-wp-widget-image.php
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	/**
+	 * Test render_media method.
+	 *
+	 * @covers WP_Widget_Image::render_media
+	 */
+	function test_render_media() {
+		$widget        = new WP_Widget_Image();
+		$attachment_id = self::factory()->attachment->create_object( DIR_TESTDATA . '/images/canola.jpg', 0, array(
```
westonruter: See note about string `$args` being deprecated.

tests/phpunit/test-class-wp-widget-media.php
```diff
@@ -0,0 +1,354 @@
...
+	 *
+	 * @covers WP_Widget_Media::create_link_for()
+	 */
+	function test_create_link_for() {
+		if ( version_compare( PHP_VERSION, '5.3', '<' ) ) {
+			$this->markTestSkipped( 'ReflectionMethod::setAccessible is only available for PHP 5.3+' );
+			return;
+		}
+
+		$attachment_id = self::factory()->attachment->create_object( DIR_TESTDATA . '/images/canola.jpg', 0, array(
```
westonruter: See note about string `$args` being deprecated.

tests/phpunit/test-class-wp-widget-media.php
```diff
@@ -0,0 +1,354 @@
...
+		// Should filter invalid key.
+		$result = $widget->update(
+			array(
+				'imaginary_key' => 'value',
+			),
+			$instance
+		);
+		$this->assertSame( $result, $instance );
+
+		add_filter( 'sanitize_text_field', array( $this, '_return_wp_error' ) );
```
westonruter, line 189: Tricky 😄

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -79,7 +79,7 @@ protected function get_instance_schema() {
 				'caption' => array(
 					'type' => 'string',
 					'default' => '',
-					'sanitize_callback' => 'sanitize_text_field',
+					'sanitize_callback' => 'wp_kses_post',
```
westonruter, line 82: Good catch!

westonruter (changes requested):

westonruter (approved):

PR #45, opened 2017-03-23
Add QUnit tests

### LINKED ISSUE #44: Add QUnit tests

westonruter:
Fixes #44


westonruter:
Woohoo! Tests pass. Not to finish fleshing out the initial tests.

westonruter:
I'm going to be largely offline for the following week. I'd be grateful for anyone who wants to work toward fleshing out the tests I started in this PR.

timmyc:
@westonruter - is this still something you could use some help with?

westonruter:
@timmyc very much so! Thanks!

timmyc:
@westonruter well I can't quite seem to figure out why the tests sometimes pass on Travis and other times do not.  They are passing locally for me each time.

I'm curious to hear your thoughts on how I'm testing the interactions with the media modal... I really wish there was a cleaner way to do this, perhaps via `set`ing model data directly on the media manager itself, but I'm not very savvy with that code.  Any ideas/feedback would be appreciated!

westonruter:
@timmyc you can see in one of the [failing builds](https://travis-ci.org/xwp/wp-core-media-widgets/jobs/219079320#L522) that the issue is there is a JSHint error:

> Missing '()' invoking a constructor.

So yeah, if you fix up the JSHint issues and then any other subsequent linting issues that get checked prior to running QUnit, the build should get fixed.

The reason why one of the build passes and the other one fails, is that the passing build for some reason isn't doing the checks. See https://github.com/xwp/wp-dev-lib/issues/197

timmyc:
Doh, I thought I had caught all those missing `()` - thanks!

westonruter:
I'm going to try to fix https://github.com/xwp/wp-dev-lib/issues/197 here. Hold off on pushing any fixes to this branch as I'm going to be pushing a commit which will be amended until I have the fix.

tests/qunit/fixtures/media-widgets.js
```diff
@@ -0,0 +1,22 @@
...
+	sizes: {
+		medium: {
+			url: 'http://src.wordpress-develop.dev/wp-content/uploads/2017/04/chicken-and-ribs-300x300.jpg'
+		}
+	},
+	title: 'Chicken and Ribs',
+	type: 'image',
+	url: 'http://src.wordpress-develop.dev/wp-content/uploads/2017/04/chicken-and-ribs.jpg',
+	width: 1080
+};
```
westonruter: I don't think you necessarily need to store this in a separate fixture file. I think it's fine to just use the literal object in the call to `imageWidgetControlInstance.selectedAttachment.set()` inline.

In addition, it would be interesting to consider overriding/mocking the `jQuery.ajax` functionality for the `get-attachment` admin ajax request to return this data instead, so that you could do `control.model.set( { attachment_id: 777 } )` and then the mock would return the expected attachment to then populate it into `selectedAttachment`.

westonruter:
Hummm. For some reason the `WP_MULTISITE=1` jobs are failing only:  https://travis-ci.org/xwp/wp-core-media-widgets/builds/219567567

> Testing tests/qunit/index.html JQMIGRATE: Migrate is installed with logging active, version 1.4.1
> .....F
> Image Media Widget - Media Modal Edits
> Message: Script error.
> Actual: null
> Expected: undefined
> :0

tests/qunit/index.html
```diff
@@ -1369,7 +1378,7 @@ <h2 class="upload-instructions drop-instructions">Drop files anywhere to upload<
 			<div class="favicon">
 				<img id="preview-favicon" src="{{ data.url }}" alt="Preview as a browser icon"/>
 			</div>
-			<span class="browser-title" aria-hidden="true">WordPress Develop</span>
+			<span class="browser-title" aria-hidden="true">WordPress Develops22s</span>
```
obenland: Is this intentional?

timmyc: I noticed that too, but when I rebuilt the test file it seemed to clear up for me.

timmyc:
@westonruter thanks for getting this green - should I build out further tests or do we want to merge this in as-is and add more tests later?

tests/qunit/index.html
westonruter: Ah, no, it's not. It's because I changed the `blogname` on my system during some tests, so that change leaked into here 😄

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -195,6 +195,19 @@
...
+			 * customizer.
+			 */
+			mediaFrameContentView = mediaFrame.views.get( '.media-frame-content' )[0];
+			if ( mediaFrameContentView.model && mediaFrameContentView.model.dfd ) {
+				mediaFrameContentView.model.dfd.done( function() {
+					_.defer( function() { // Next tick.
+						mediaFrameContentView.$el.find( '[data-setting="caption"]:first' ).focus();
+					} );
+				} );
+			}
```
westonruter: This chunk I added seems now out of place for this PR and it could be good to remove. Auto-focusing was addressed in another PR and it also seems like an issue the media JS should tackle anyway.

westonruter:
> thanks for getting this green - should I build out further tests or do we want to merge this in as-is and add more tests later?

@timmyc I want to try to restore some of your code that I removed.

It'll be interesting to see how the code will be merged into core, especially the WP-CLI command for regenerating the test suite file, since that hasn't been done in core before.

timmyc:
> especially the WP-CLI command for regenerating the test suite file, since that hasn't been done in core before.

I thought that was pretty slick myself... was very intuitive and easy to figure out just based off the comment in the `index.html` file for me.  Want me to add the tests/more tests back in now or are you still working on this branch?

westonruter:
@timmyc I'm not working on it at the moment. I was wrestling with opening the media modal in the tests, which led me to think that maybe we should not do that from unit testing but rather only directly set the `control.model`'s attributes instead. Unless we can figure out how to get the media JS to play nicely with qunit, testing interactions with the media modal may be better left to acceptance tests.

timmyc:
@westonruter I'm hitting similar issues with Travis and adding in new tests that rely upon `MediaFrame` interactions... even just manually setting model attributes on the media frame.  You can see the errors in Travis:

`mediaFrame.content.get( '.attachments-browser' ).sidebar.get`

Perhaps we could refactor some of these methods like `getSelectFrameProps` and subsequently `_getAttachmentProps` to not call out to mediaFrame elements directly and make them a bit easier to test using plain objects?  Still somewhere there will have to be code that grabs various aspects of `state` from the `MediaFrame`.

Again these tests pass fine locally, but Travis I can't seem to find the secret `setTimeout` magic to make it work... which is a bit of a code smell to me that the tests are that fragile.

Thoughts?

westonruter:
@timmyc the tests aren't passing for me locally actually.

I agree that the `setTimeout` feels fragile. I think your idea to split out the logic to make them easier to test independently of the media JS is a wise way to go.

Aside, I found that opening the `index.html` in Chrome is way better for debugging than looking at CLI output:

image

westonruter:
@timmyc yeah, let's definitely not try to invoke any media JS in QUnit testing. See https://wordpress.slack.com/archives/C02SX62S6/p1491938808087497

timmyc:
> https://wordpress.slack.com/archives/C02SX62S6/p1491938808087497

Glad to read the confirmation there 😆 

For the sake of moving this PR forward, I will rip out all of the mediaFrame centric tests and get Travis back to 💚 .  My vote is at that point to get this PR merged in, so we can then refactor a bit to make some of the code a bit easier to test.  Also just getting this branch in will allow us to build out even basic coverage for Video and Audio widgets too.

timmyc (approved):
🚢  looks good to me

Issue #9, opened 2017-03-09
Allow adding media by external URL


obenland:
Add  the menu so media can be added via URL.

See https://core.trac.wordpress.org/ticket/32417#comment:69

obenland:
This is not a PR because I don't know enough about the media modal API (yet) to make it happen

westonruter:
Essentially we need to add the “Insert from URL” tab:

![image](https://cloud.githubusercontent.com/assets/134745/23982584/8ef0d786-09cb-11e7-8036-ce1bb8afceda.png)

westonruter:
Some todos/questions I see here:

* We need to open the media selection frame with the new “Insert from URL” tab shown, in addition to adding a “Insert ~Media~Image” tab for the media library selection frame.
* When an external image has been selected, the Edit Image button should open to the “Insert from URL” view with the `url`. I don't think the edit image modal shown for attachments will work.
* While the media manager is anticipating that you can paste any media URL in, we need to constrain it specifically for images.
* An image added by external URL will result in the widget instance having an `attachment_id` of `0` and yet a `url` pointing to the external image.
* The image preview JS template in `\WP_Widget_Image::render_control_template_scripts()` needs to account for the case when `url` is populated while `attachment_id` is `0`.
* Similarly, the `\WP_Widget_Image::render_media()` method needs to be updated to account for when `$instance['attachment_id']` is empty while `$instance['url']` is populated.

obenland:
Great summary, thanks for that!

> in addition to adding a “Insert MediaImage” tab for the media library selection frame.

I  think it's fine for  it to remain to say "Insert Media", we can still filter the actual library.

>  I don't think the edit image modal shown for attachments will work.

It looks like the editor uses the edit image modal, it just doesn't show size or image editing options and you can't link it to its attachment page. I wonder if the modal handles that case out of the box based on whether an attachment id is provided or not? Worth trying out.

adamsilverstein:
I'm going to take a look at this issue

adamsilverstein:
I started working on this in https://github.com/xwp/wp-core-media-widgets/compare/master...adamsilverstein:issue-9?expand=1 and have the insert from url part of the modal showing so far.

Next, I will work on handling the insert button, the templating and also what happens when you re-open the dialog as Weston outlined above.

adamsilverstein:
[...]

PR #46, opened 2017-03-26
Enable embedding images by URL

adamsilverstein:
Addresses https://github.com/xwp/wp-core-media-widgets/issues/9

* Add a custom Post modal state to handle our specific needs (select/embed)
* Modify the insert button language temporarily - cleanest approach for insert button. I tried overriding the setup methods and that was messier. 
* restore standard post state and language on close
* handle embed data on close
* Extend the template to handle embedded images

Q: Not sure about the edit dialog - or the edit details sidebar in the media select mode: do we really want all of these attributes editable? Maybe the widget itself has a title and description and the media part is only image selection? I'm confused about how these would be used - related: https://core.trac.wordpress.org/ticket/32417#comment:132


melchoyce:
Looking good!

Some things I noticed:

If you go back and edit an image after picking it for the widget (for example, cropping it, or flipping it horizontally), the updated image appears in your site preview, but not in the widget preview:

screen shot 2017-03-27 at 1 45 40 pm

Also, when you change images, the image settings (size, link, etc.) revert to default. I just noticed that you also can't edit these when selecting or changing an image, only when editing. You should be able to change these settings while you're picking an image, IMO.

obenland:
When updating an URL-image I get an undefined error  for [`mediaFrameContentView.model.dfd`](https://github.com/xwp/wp-core-media-widgets/blob/c47c013e206726d79b9024d320b56022e27bc0e6/wp-admin/js/widgets/media-image-widget.js#L148)

adamsilverstein:
@melchoyce @obenland  - thanks for the review and feedback, I'll work on addressing the issues you mentioned.

adamsilverstein:
@obenland I addressed the console error you were seeing in https://github.com/xwp/wp-core-media-widgets/pull/46/commits/047e7e62a2a5a6b8cff930498b1380fb4c929043

adamsilverstein:
With the latest commits, I have completed support for embedding remote images via URL. This should be close to or ready for merge, the basic functionality is in place.

In my testing, all of the fields including the caption, alignment, and the advanced link and class options all work correctly for embedded by url media.

These fields:



all render correctly:



I'd like to address the issues [@melchoyce raised](https://github.com/xwp/wp-core-media-widgets/pull/46#issuecomment-289530257) on a separate issue.

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -43,33 +43,28 @@
...
+			if ( ! _.isEmpty( attachment ) ) {
+				props = {
+					attachment_id: attachment.id,
+					alt: attachment.alt,
+					caption: attachment.caption,
+					image_classes: '',
+					image_title: '',
+					link_classes: '',
+					link_rel: '',
+					link_url: attachment.link,
```
westonruter: Shouldn't this be `attachment.linkUrl`?

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -147,11 +142,17 @@
...
+					_.defer( function() { // Next tick.
+						mediaFrameContentView.$el.find( '[data-setting="caption"]:first' ).focus();
+					} );
 				} );
-			} );
+			} else {
+				_.defer( function() {
+					jQuery( '.media-frame-content [data-setting="caption"]:first' ).focus();
+				} );
+			}
```
westonruter: We could also just remove all of this autofocus logic from the widget itself. To me it seems like this should actually be handled in the media library JS itself, right? I mean, right now it seems that the [`this.$el.focus();` call](https://github.com/WordPress/wordpress-develop/blob/4.7.3/src/wp-includes/js/media-views.js#L6805) in `Modal#focus()` should perhaps be improved to automatically autofocus on the first frame's first focusable element?

wp-admin/js/widgets/media-widgets.js
```diff
@@ -1,5 +1,5 @@
 /* eslint consistent-this: [ "error", "control" ] */
-wp.mediaWidgets = ( function( $ ) {
+wp.mediaWidgets = ( function( $, _wpMediaViewsL10n ) {
```
westonruter: Instead of accessing this “private” global variable, should we not instead access `wp.media.view.l10n`?

wp-admin/js/widgets/media-widgets.js
```diff
@@ -141,6 +236,21 @@ wp.mediaWidgets = ( function( $ ) {
...
 			var control = this, attachment;
 
+			// This is an embed (by URL) image if the url is set and the attachment_id is 0.
+			if ( 0 === control.model.get( 'attachment_id' ) && control.model.get( 'url' ) ) {
+
+				// Construct an attachment model with the data we have.
+				attachment = new wp.media.model.Attachment( control.model.attributes );
+
+				// Once the widget renders, inject the preview.
+				_.defer( function() {
```
westonruter: This `defer` seems a bit brittle and it may indicate we should improve something with how `selectedAttachment` is used.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -131,6 +132,100 @@ wp.mediaWidgets = ( function( $ ) {
...
+					return display;
+				}
+			} );
+
+			/**
+			 * Extend wp.media.view.MediaFrame.Post for our simplified media upload modal.
+			 */
+			control.originalButtonLanguage = _wpMediaViewsL10n.insertIntoPost;
+			control.originalMediaFramePost = wp.media.view.MediaFrame.Post;
+			control.customMediaFramePost = wp.media.view.MediaFrame.Post.extend( {
```
westonruter: I think this should be `CustomMediaFramePost` since it is a class.

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -148,49 +148,69 @@ public function render_media( $instance ) {
...
+			$alt     = isset( $instance['alt'] ) ? $instance['alt'] : '';
+			$align   = isset( $instance['align'] ) ? $instance['align'] : '';
+			$width   = isset( $instance['width'] ) ? $instance['width'] : '';
+			$height  = isset( $instance['height'] ) ? $instance['height'] : '';
+			$image = do_shortcode(
+				sprintf(
+					'[caption id="0" align="%s" width="%d"]<img class="" src="%s" alt="%s" width="%d" height="%d" />%s[/caption]',
+					$align,
+					$width,
+					$url,
```
westonruter: Add `esc_url()` for good measure?

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -148,49 +148,69 @@ public function render_media( $instance ) {
...
+					$alt,
```
westonruter: Add `esc_attr()`?

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -148,49 +148,69 @@ public function render_media( $instance ) {
...
+			$url     = $instance['url'];
+			$caption = isset( $instance['caption'] ) ? $instance['caption'] : '';
+			$alt     = isset( $instance['alt'] ) ? $instance['alt'] : '';
+			$align   = isset( $instance['align'] ) ? $instance['align'] : '';
+			$width   = isset( $instance['width'] ) ? $instance['width'] : '';
+			$height  = isset( $instance['height'] ) ? $instance['height'] : '';
+			$image = do_shortcode(
+				sprintf(
+					'[caption id="0" align="%s" width="%d"]<img class="" src="%s" alt="%s" width="%d" height="%d" />%s[/caption]',
+					$align,
```
westonruter: Add `esc_attr()` for crossing i's and dotting t's?

wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -148,49 +148,69 @@ public function render_media( $instance ) {
...
+			// @todo Add caption and other embedded data handling.
+			$url     = $instance['url'];
+			$caption = isset( $instance['caption'] ) ? $instance['caption'] : '';
+			$alt     = isset( $instance['alt'] ) ? $instance['alt'] : '';
+			$align   = isset( $instance['align'] ) ? $instance['align'] : '';
+			$width   = isset( $instance['width'] ) ? $instance['width'] : '';
+			$height  = isset( $instance['height'] ) ? $instance['height'] : '';
+			$image = do_shortcode(
+				sprintf(
+					'[caption id="0" align="%s" width="%d"]<img class="" src="%s" alt="%s" width="%d" height="%d" />%s[/caption]',
```
westonruter: Why not just call `img_caption_shortcode()` directly?

westonruter (changes requested):

adamsilverstein:
@westonruter thanks for the review and feedback!

obenland:
> Q: Not sure about the edit dialog - or the edit details sidebar in the media select mode: do we really want all of these attributes editable?

I'm not sure how I feel about them being removed. Without the controls it looks like any attached image gets automatically liked to its attachment page. It's also inconsistent with the embed modal where it still lets you align images and add a custom link url.
While I understand that aligning an image within a widget probably doesn't make much sense, the other attributes feel like they still have a place there.

@westonruter With the exception of https://github.com/xwp/wp-core-media-widgets/pull/46#discussion_r109343180 and https://github.com/xwp/wp-core-media-widgets/pull/46#discussion_r109343522 your feedback should be addressed.

obenland:
One bug that I noticed: When pasting a non-image url, like an audio file, it returns the most recently uploaded picture

adamsilverstein:
@obenland thanks for your work here - i can look at the remaining issues.

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -69,6 +88,38 @@
...
+				props = {
+					attachment_id: attachment.id,
+					align: attachment.align,
+					alt: attachment.alt,
+					caption: attachment.caption,
+					image_classes: '',
+					image_title: '',
+					link_classes: '',
+					link_rel: '',
+					link_url: attachment.linkUrl ? attachment.linkUrl : attachment.link,
```
westonruter: Is `attachment.link` a URL when `attachment.linkUrl` is empty?

adamsilverstein: I thought link was the link type (self, attachment page, other) and the linkUrl was the actual URL. I will test this out.

wp-admin/js/widgets/media-image-widget.js
adamsilverstein: I'm happy to remove this logic here and agree the media modal should be taking care of the focus. It does this already focusing on the modal itself, which may be enough.

wp-admin/js/widgets/media-image-widget.js
adamsilverstein: here is the passed data when I edit an embedded image:

wp-admin/js/widgets/media-widgets.js
adamsilverstein: Couldn't quite track down why this is required - without it, the embedded items are set immediately and never rendered. It may be the sequence of event handlers and how they are added. We can dig into this more post merge.

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -106,7 +106,7 @@
 					image_title: '',
 					link_classes: '',
 					link_rel: '',
-					link_url: attachment.linkUrl ? attachment.linkUrl : attachment.link,
+					link_url: attachment.linkUrl,
```
obenland: `linkUrl` is only set when it's a custom link. When it's linked to the image, it doesn't pass the image url again.
So it looks like we need to be more verbose here and check for what's in `link` and populate `link_url` accordingly.

westonruter: ¿So like:

```js
link_url: 'custom' === attachment.link ? attachment.linkUrl : '',
```

obenland: I currently have 
```javascript
if ( 'custom' === attachment.link ) {
	linkUrl = attachment.linkUrl;
} else if ( 'file' === attachment.link ) {
	linkUrl = attachment.url;
}
```
on my sandbox.

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -187,23 +187,6 @@
...
-				mediaFrameContentView.model.dfd.done( function() {
-					_.defer( function() { // Next tick.
-						mediaFrameContentView.$el.find( '[data-setting="caption"]:first' ).focus();
-					} );
-				} );
-			} else {
-				_.defer( function() {
-					jQuery( '.media-frame-content [data-setting="caption"]:first' ).focus();
-				} );
-			}
```
westonruter: @adamsilverstein do you think that the improvements to media modal focusing will be made to the underlying media JS? Do you concur that there seems to be a problem with the focus behavior when opening the modal?

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -98,7 +98,7 @@
 
 			if ( ! _.isEmpty( attachment ) ) {
 				props = {
-					attachment_id: attachment.id,
+					attachment_id: 0,
```
obenland: This causes my control preview to not update some times. But not always. Still trying to work out why. Does this work for everyone else?

wp-admin/js/widgets/media-widgets.js
```diff
@@ -397,6 +397,7 @@ wp.mediaWidgets = ( function( $ ) {
 
 				// Update widget instance.
 				control.model.set( control.getSelectFrameProps( mediaFrame ) );
+				control.model.trigger( 'update' );
```
westonruter: Why a custom `update` event here? Wouldn't it suffice to just change the above `change:attachment_id` to just `change` and then this custom event trigger could be removed?

westonruter: The `update` event is supposed to be for collections:
> single event triggered after any number of models have been added or removed from a collection.

wp-admin/js/widgets/media-image-widget.js
```diff
@@ -43,32 +43,78 @@
...
+		 *
+		 * @param {object} attachment Attachment object.
+		 * @returns {object} Props
+		 */
+		_getEmbedProps: function _getEmbedProps( attachment ) {
+			var props = {};
+
+			if ( ! _.isEmpty( attachment ) ) {
+				props = {
+					attachment_id: 0,
```
westonruter: Why is this `attachment_id: 0` and not `attachment_id: attachment.id`?

westonruter:
@melchoyce the changes here remove the Attachment Display Settings from the select frame:

image

Is that OK?

wp-admin/js/widgets/media-image-widget.js
westonruter: Yes, `link` is the type and `linkUrl` is the actual URL. So I don't think that the type (`attachment.link`) should ever be made the `link_url`.

wp-admin/js/widgets/media-image-widget.js
westonruter: @obenland that should work, but it would seem to be somewhat redundant. If the `link_type` is `file` then the PHP logic isn't going to look at `link_url` anyway, and will instead look at `url` anyway, per `\WP_Widget_Image::render_media()`:

```php
$url = '';
if ( 'file' === $instance['link_type'] ) {
	$url = $attachment ? wp_get_attachment_url( $attachment->ID ) : $instance['url'];
} elseif ( $attachment && 'post' === $instance['link_type'] ) {
	$url = get_attachment_link( $attachment->ID );
} elseif ( 'custom' === $instance['link_type'] && ! empty( $instance['link_url'] ) ) {
	$url = $instance['link_url'];
}
```

melchoyce:
Why are we removing attachments? Being able to link your image somewhere seems very useful.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -258,95 +368,36 @@ wp.mediaWidgets = ( function( $ ) {
...
+			mediaFrame.on( 'close', function() {
+				var attachment, state = mediaFrame.state();
+
+				// Restore the original wp.media.view.MediaFrame.Post object and language.
+				wp.media.view.MediaFrame.Post = control.originalMediaFramePost;
+				wp.media.view.l10n.insertIntoPost = control.originalButtonLanguage;
 
 				// Update cached attachment object to avoid having to re-fetch. This also triggers re-rendering of preview.
-				attachment = mediaFrame.state().get( 'selection' ).first().toJSON();
+				attachment = 'embed' === state.get( 'id' ) ? state.props.toJSON() : state.get( 'selection' ).first().toJSON();
```
westonruter: I was getting as JS error here, though I'm trying to figure out how to reproduce:

image

westonruter: It happens if I open the media modal and then close it without having selected anything.

wp-admin/js/widgets/media-image-widget.js
westonruter: I see. The `attachment.id` is retained in `ImageWidgetModel#_getAttachmentProps` but is not relevant here in `ImageWidgetModel#_getEmbedProps` since there is no underlying attachment.

westonruter:
> Why are we removing attachments? Being able to link your image somewhere seems very useful.

@melchoyce attachments aren't being removed. It's just the Attachment _Display Settings_ fields on the select frame.

wp-admin/js/widgets/media-widgets.js
westonruter: It's because the event was changed from `select` to `close`. Working on a fix.

melchoyce:
But then how would you add a link?

melchoyce:
Testing — can't figure out how to add a link as I'm adding an image. Also noticed we're defaulting to thumbnail, which crops the image:

screen shot 2017-04-04 at 3 41 46 pm

It's confusing to see one thing in the widget and another in the preview. They should match. 

Can we default to a not-cropped version?

wp-admin/js/widgets/media-widgets.js
westonruter: Fixed in d284087.

melchoyce:
After testing this against the master branch right now, I definitely prefer the defaults in the master branch.

wp-admin/js/widgets/media-image-widget.js
westonruter: I think this may be fixed as of 8c3ae25

adamsilverstein:
Did we loose the image settings from the regular image attachment frame as well? if so that is a mistake and I will fix that.

When selecting an embed image, you can't select an image size because we aren't importing and resizing the image, we are using it as is. i think you can add a link as you add the image?

adamsilverstein:
I see, linking is available from the 'insert from url' flow, but it got lost on the insert from media state. i will fix this @melchoyce and test against master again, I agree it should be available while you are adding the image.

Insert from URL

wp-admin/js/widgets/media-widgets.js
westonruter: @adamsilverstein changed in 8c3ae2581cf580

wp-admin/js/widgets/media-widgets.js
westonruter: Fixed in 0b84597

westonruter:
> I see, linking is available from the 'insert from url' flow, but it got lost on the insert from media state. i will fix this @melchoyce and test against master again, I agree it should be available while you are adding the image.

@adamsilverstein @melchoyce I believe I've fixed this in https://github.com/xwp/wp-core-media-widgets/pull/46/files/0b84597..1ca00e5

westonruter (approved):

wp-admin/js/widgets/media-widgets.js
```diff
@@ -200,7 +204,7 @@ wp.mediaWidgets = ( function( $ ) {
 				 *
 				 * @return {void}
 				 */
-				createStates: function() {
+				createStates: function createStates() {
```
adamsilverstein: @westonruter can you explain the advantages or why you prefer this pattern as opposed to 'anonymous' functions?

wp-admin/js/widgets/media-widgets.js
```diff
@@ -370,6 +367,7 @@ wp.mediaWidgets = ( function( $ ) {
 
 			if ( ! control.templateRendered ) {
 				control.$el.html( control.template()( control.model.attributes ) );
+				control.renderPreview(); // Hereafter it will re-render when control.selectedAttachment changes.
```
adamsilverstein, line 374: @westonruter Brilliant, thank you! I was struggling to find where this re-rendering was failing, this makes way more sense.

adamsilverstein:
I did some testing with the latest code and everything worked well, going to see if I can address the cropping/rotation in the preview.

adamsilverstein:
@melchoyce I fixed the issue with image cropping/rotation not showing in the preview...

In https://github.com/xwp/wp-core-media-widgets/pull/46/commits/36fc72588a56e3339d83d061ac2d6dee1afe8fb5 I switched the preview template to using only the image 'url' property for display. WordPress changes this data point when you rotate or crop an image, causing the preview to re-render correctly. This simplifies the template and we can control the preview size with CSS.

melchoyce:
Awesome! Need me to test?

westonruter:
@melchoyce yes, please test and once you verify we'll merge.

wp-admin/js/widgets/media-widgets.js
westonruter: Yes. The reason for naming the functions is for stack traces. Compare:

image

More reasons here: https://github.com/getify/You-Dont-Know-JS/blob/master/scope%20%26%20closures/ch3.md#anonymous-vs-named

melchoyce (approved):

melchoyce:
🎉

PR #52, opened 2017-04-06
Prevent rendering widget if instance lacks valid attachment_id or url

westonruter:


wp-includes/widgets/class-wp-widget-image.php
```diff
@@ -150,7 +150,10 @@ public function render_media( $instance ) {
 			'size' => 'thumbnail',
 		) );
 
-		$attachment = get_post( $instance['attachment_id'] );
+		$attachment = null;
+		if ( $instance['attachment_id'] ) {
+			$attachment = get_post( $instance['attachment_id'] );
+		}
```
westonruter, line 156: Adding this condition was key because if `attachment_id` is `0` then `get_post( 0 )` will get whatever the global `$post` is.

obenland: That's not good!

obenland (approved):

PR #53, opened 2017-04-07
Add media-video-widget

### LINKED ISSUE #32: Implement video widget

timmyc:
Closes #32

Still very much a work-in-progress

TODO
- [ ] Add qunit tests
- [x] Add phpunit test coverage
- [x] Fix preview rendering
- [x] Add support for poster images (`select-poster-image`).
- [x] Add support for alternative video sources (`add-video-source`).
- [x] Fix assigning tracks to video (`add-track`).
- [x] Placeholder image to be shown while poster image loads
- [x] Fallback icon/image for external videos other than YouTube and Vimeo
- [x] Ensure that all media JS/CSS get enqueued on the frontend in the customizer preview, including `froogaloop` for Vimeo (see `wp_video_shortcode()`).
- [x] Show preview of embed in “Insert from URL” frame using noembed.com while waiting for oEmbed proxy endpoint ([#40450](https://core.trac.wordpress.org/ticket/40450)), by overriding `EmbedLink#fetch` to source the data from the endpoint instead of from the `parse-embed` admin-ajax request.
- [ ] ~Update oEmbed previews once [#40450](https://core.trac.wordpress.org/ticket/40450) lands to use oEmbed proxy endpoint, though core would likely make use of it as well (so plugin may continue with noembed.com polyfill). See https://github.com/xwp/wordpress-develop/pull/225~ See #94.
- [ ] ~Prevent re-rendering video preview when none of the fields have changed, e.g. when changing the title.~ See #92
- [ ] ~Ensure that `width`/`height` are handled properly to not overflow widget container, especially when ME.js is disabled via `add_filter( 'wp_video_shortcode_library', '__return_empty_string' );`~. See #93.
- [ ] ~Improve rendering of `poster` images when ME.js will show YouTube or Vimeo players with their own poster frames underneath.~ See #95.


timmyc:
Okay I have the `renderPreview` working akin to how it is done within the editor using the `parse-media-shortcode` admin-ajax action to fetch iframe contents.  This would be a neat place to use the REST API too.  Still some formatting and general polish needed, but progress:

customize__blog_ _wordpress_develop_

westonruter:
@timmyc something else to consider is opting for using JS rendering of the video when `attachment_id` is populated, and it's not rendering an oEmbed video. For example, from an earleir version of this plugin:

https://github.com/xwp/wp-core-media-widgets/blob/5e257adff436ee831df9fe3b58d07673da8703a0/wp-media-widget.js#L300-L328
https://github.com/xwp/wp-core-media-widgets/blob/5e257adff436ee831df9fe3b58d07673da8703a0/class-wp-widget-media.php#L395-L397

Is that viable? We only need the preview in the widget control to just _show_ the video that is selected. We don't need to worry about the `height`, `width`, and any other properties that the user may supply because these will be previewed in the customizer preview which will render the chosen video in high fidelity. 

To me it makes sense to use the admin-ajax approach to render the video only in the case of an _external video_ URL (when instance `url` is populated). 

Nevertheless, and maybe this isn't as accessible, in that case perhaps we should just show an informational message or placeholder? If we take the example of the header video feature in 4.7, there is only an input field for taking the video URL:

image

I think we can greatly simplify this if we take a similar basic approach.

For YouTube videos, we could actually parse the poster frame image and show that in the widget when rendering the preview. For example:

Given video URL: https://www.youtube.com/watch?v=KiS8rZBeIO0
=>
Poster image is: https://i1.ytimg.com/vi/KiS8rZBeIO0/hqdefault.jpg

There is a similar facility for Vimeo, although it requires making an API call unfortunately: http://stackoverflow.com/a/1361192/93579

@melchoyce @obenland thoughts?

timmyc:
@westonruter I do like that approach for rendering the video - I wasn't aware that was an option tbh - but did like the `admin-ajax` approach since it _seemed_ it would work with external services as well, but avoiding iframes is always 💯 in my book.

> If we take the example of the header video feature in 4.7, there is only an input field for taking the video URL

I think this is a great idea to borrow the existing patterns from the video header widget whenever possible.

melchoyce:
Not a fan of the extra field for URL embedding in general — what happens if you add a video to both? It's not clear — so I'd prefer if we could embed videos through "Insert from URL" ("Insert from Youtube?") in the media library:

screen shot 2017-04-11 at 10 23 18 am

I think I'll make a core ticket to address this as well. 

(FWIW, if you add a video URL, and you already have a video uploaded, it doesn't show at all.)

melchoyce:
> For YouTube videos, we could actually parse the poster frame image and show that in the widget when rendering the preview. 

I like this idea 👍

timmyc:
In f9cf4de I have rebased and fixed the rendering for external videos.  I explored the `embed` route using admin-ajax as well.  Looks like that route might not work due to the various admin-ajax actions returning when a `post_ID` is not supplied on the request.

I dug in a bit more to how the Header Video handles the preview, and it renders the video markup and creates a new `MediaElementPlayer`[  [#](https://core.trac.wordpress.org/browser/tags/4.7.3/src/wp-admin/js/customize-controls.js#L2499) ].  This only works on locally hosted video - there is no preview for YouTube URLs within the widget control - only the customizer preview itself.

As such I will explore [this approach](https://github.com/xwp/wp-core-media-widgets/pull/53#issuecomment-292682321).  Do we think it is okay to just show a link to the video file when a local media item is selected?  Or should we perhaps use the `MediaElementPlayer` approach for local media?

melchoyce:
> Do we think it is okay to just show a link to the video file when a local media item is selected? Or should we perhaps use the MediaElementPlayer approach for local media?

Does this mean if you upload a video to your media library? We should always embed the video onto your site, whether your video is coming from YouTube, or from your media library.

timmyc:
> Does this mean if you upload a video to your media library? We should always embed the video onto your site, wether your video is coming from YouTube, or from your media library.

Sorry @melchoyce for the confusion in my last comment.  I'm only speaking to the rendering of the preview itself in the context of the customizer sidebar, or the widget admin.  The current route I took to handle the preview render was to use `admin-ajax` and grab the embed markup and render in an iframe - the same way it is handled when showing a video preview in the editor.  This works decently for locally hosted files - the iframe renders the video as expected, but when trying the same route for a generic `embed` like YouTube, or Vimeo, it fails since the admin-ajax logic is dependent on a post ID being passed in.

It looks like in one of the prior iterations on the video widget, that @westonruter linked to above ( https://github.com/xwp/wp-core-media-widgets/blob/5e257adff436ee831df9fe3b58d07673da8703a0/wp-media-widget.js#L300-L328 ) - The preview was rendering a simple <a> link for locally hosted media, and a video player for external files... though I haven't tested this out yet.

My inclination is to try to show a video player in the Widget Preview regardless of the source.  I understand while using the widget within the customizer, the widget will properly render the video in the customizer preview properly, but if the widget is being managed via the widget admin interface only - just showing a link to the local video file doesn't feel like a great UX to me.  The fact the Header Video doesn't show a preview for a YouTube video _feels_ a bit broken to me... just my .02 though.

timmyc:
Okay here are some quick screen grabs of how things look when attempting to use the `admin-ajax` preview:

__Video From Media Library__

Note the lovely horizontal scroll on the iframe.  But live preview inside the customizer is looking as expected

__Video from YouTube__
customize__blog_ _wordpress_develop 3

Embed preview fails since no post_ID is passed to the admin-ajax call, but customizer preview is looking good

westonruter:
@timmyc if we switch to using the poster frame from YouTube instead of try to embed the video via `admin-ajax`, then that would fix this:

image

timmyc:
> if we switch to using the poster frame from YouTube instead of try to embed the video via admin-ajax, then that would fix this:

@westonruter but should we aim for a solution that isn't 3rd party dependent?  Seems like if we could leverage oembeds at least we don't have to worry about differences between platforms like YouTube vs Vimeo etc.

westonruter:
@timmyc what if instead of using the admin-ajax shortcode parser, we instead just use the oEmbed protocol directly? I don't think there is currently an admin-ajax endpoint for parsing oEmbeds, but if there was one we could then provide a URL and get back the `thumbnail_url` according to the [oEmbed spec](http://oembed.com/). 

A Vimeo URL: https://vimeo.com/108650530
oEmbed endpoint: https://vimeo.com/api/oembed.json?url=https%3A%2F%2Fvimeo.com%2F108650530
`thumbnail_url`:


A YouTube URL: https://www.youtube.com/watch?v=xrGAQCq9BMU
oEmbed endpoint: https://www.youtube.com/oembed?url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DxrGAQCq9BMU&format=json
`thumbnail_url`:

westonruter:
An oEmbed endpoint, via the REST API or via admin-ajax, would essentially be implementing the guts of `\WP_oEmbed::get_html()`, minus the `data2html()` call at the end: https://github.com/WordPress/wordpress-develop/blob/4.7/src/wp-includes/class-oembed.php#L333-L360

It looks like the “Insert from URL” frame is also failing to show a normal preview due to the `post_ID` being `0`:

image

timmyc:
> It looks like the “Insert from URL” frame is also failing to show a normal preview due to the post_ID being

I had just noticed that too 😢 - wondering _why_ the `post_ID` is a hard requirement here, when we can lean on the shortcode logic to `render_media`.

So using the [prior approach](https://github.com/xwp/wp-core-media-widgets/pull/53#issuecomment-292682321) in 6b370db
 via `wp_underscore_video_template` results in the following:

__Local Video__
Much nicer experience, no iFrame, fast render:

__YouTube__
Player rendered, but using a YouTube embed url as a src for `video` obviously doesn't work:

It does seem like utilizing stills from the video providers might be a good way to solve this.  We would still need to have some sort of fallback state where perhaps we show a generic video icon that is linked to the external site?

timmyc:
And travis is acting quite strange with this branch...  I did a rebase, but it is saying no custom travis config is found so it then attempts to run `rake` against the project 😆 

xwp_wp-core-media-widgets_-_travis_ci

westonruter:
There is actually a REST API for `embed`, although currently it is limited to handling (post) URLs on the site. I've raised a question about opening this for handling external providers: https://wordpress.slack.com/archives/C02RQC26G/p1492038272597233

> It does seem like utilizing stills from the video providers might be a good way to solve this. We would still need to have some sort of fallback state where perhaps we show a generic video icon that is linked to the external site?

Yes. A generic video icon should be shown in that case.

timmyc:
Added in slack too, but for discussion sake here - the similar embed endpoint in the wpcom API:  https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/embeds/render/

Given that an existing endpoint does not exist - do we consider that a blocker or look at possible other solutions for now:

- Look at adding support in the `admin-ajax` call to allow responding to requests without a `post_ID` - which would also fix the selection screen preview
- Cobble together solutions for YouTube preview thumbnails since it has a standard pattern based off video id
- And add the fallback "generic" icon / link for all external videos for now

One other glitch with the external URLs in the current state, is there is no way to edit the URL once entered.  You have to remove the widget and start over currently.

westonruter:
> Look at adding support in the admin-ajax call to allow responding to requests without a post_ID - which would also fix the selection screen preview

I think the reason for why `post_ID` is required is because currently WP will cache oEmbed responses in postmeta: https://github.com/WordPress/wordpress-develop/blob/4.7/src/wp-includes/class-wp-embed.php#L258-L264

westonruter:
> Cobble together solutions for YouTube preview thumbnails since it has a standard pattern based off video id

This would be an option for YouTube, and for Vimeo we'd be able to lookup the `thumbnail_id` via oEmbed thanks to their response allowing CORS. It would be a shame to have different solutions for each provider. 

One option is <https://noembed.com/>. It actually supports CORS and provides a unified interface. So we could do XHR requests for:

https://noembed.com/embed?url=https://vimeo.com/108650530
https://noembed.com/embed?url=https://www.youtube.com/watch?v=KiS8rZBeIO0

And in both cases we could get the the `thumbnail_id` out of it. This would be an alternative to building the same kind of proxy into WordPress's own `oembed` REST API endpoint.

westonruter:
CORS requests are supported by IE8 (under a different name, `XDomainRequest`): https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Browser_compatibility

So this would be a suitable approach in terms of browser-support.

timmyc:
> It would be a shame to have different solutions for each provider.

➕ so much complexity for what seems to be such a simple problem.  If the header video is any indication though, that has been the route to take thus far with this problem.  Though not sure why they didn't try the shortcode approach there for front-end rendering... maybe I need to get in touch with the devs of that feature and see if there is something we might be missing here. 🤔 

> One option is https://noembed.com/. 

Would this be a logical option to ship to such a large portion of the web?  Granted this service wouldn't be getting hit on every page load, but still introducing a 3rd party dependency like this kind of makes me pause a bit.

timmyc:
In 2b5ccfd I have added in thumbnails for external services.  Currently YouTube is generated by plucking the video ID from the URL.  They do not support CORS afaik unless you use their API which needs a key.

Vimeo thumbnail is fetched via a CORS request.  I haven't tested this on IE8 but it is using jQuery's ajax to perform the operation so hopefully that will work properly in IE8 as well.

The fallback thumbnail/icon still needs to be added, but it is stubbed out.  All thumbnails could possibly use a loading state to prevent the widget from resizing while the image is fetched/rendered.

thumbs

westonruter:
Nice work. I think that will serve us well while we wait to figure the REST API `embed` endpoint.

wp-admin/js/widgets/media-video-widget.js
```diff
@@ -0,0 +1,236 @@
...
+		 * @returns {void}
+		 */
+		editMedia: function editMedia() {
+			var control = this, mediaFrame, metadata, updateCallback;
+
+			metadata = {
+				attachment_id: control.model.get( 'attachment_id' ),
+				caption: control.model.get( 'caption' ),
+				description: control.model.get( 'description' ),
+				link_type: control.model.get( 'link_type' )
```
westonruter: There are missing properties being passed from the model into the media frame. For example, `autoplay` and `loop`.

westonruter:
> And travis is acting quite strange with this branch... I did a rebase, but it is saying no custom travis config is found so it then attempts to run rake against the project 😆

Incidentally, this seems to have fixed itself. 🎉

wp-includes/widgets/class-wp-widget-media-video.php
```diff
@@ -70,6 +70,12 @@ public function get_instance_schema() {
 					'default' => '',
 					'sanitize_callback' => 'wp_kses_post',
 				),
+				'poster' => array(
```
timmyc: This attribute name confused me at first ( could be lack of 2nd cup of ☕️  ) - at first glance I thought it was related to the original user that posted the video... the description and usage in the code cleared it up for me though.

timmyc: And this https://codex.wordpress.org/Function_Reference/wp_video_shortcode#Attributes clarifies it further.  _goes for 2nd cup of_ ☕️

wp-admin/js/widgets/media-video-widget.js
```diff
@@ -68,47 +68,44 @@
...
-		 * @returns {Promise} CORS response
+		 * @returns {Promise} Promise that resolves with oEmbed object containing thumbnail_url.
 		 */
 		getExternalThumbnail: function getExternalThumbnail() {
-			var control = this, url, isYouTube, isVimeo, deffered;
-			url = control.model.get( 'url' );
-			isYouTube = url.match( /youtube|youtu\.be/ );
-			isVimeo = -1 !== url.indexOf('vimeo');
-			deffered = $.Deferred();
+			var control = this, urlParser = document.createElement( 'a' );
```
timmyc: Great refactor here, thanks!  I had grabbed much of the code above from `wp_underscore_video_template`, but really love how this method reads now, nice work 🌟

westonruter:
- [ ] Show preview of embed in “Insert from URL” frame. At least a poster frame should be shown while we wait for an embed proxy endpoint. Once there is an embed proxy endpoint, we might be able to override `EmbedLink#fetch` to source the data from the endpoint instead of from the `parse-embed` admin-ajax request, though presumably this would be done in core as well.
- [x] Add `target=_blank` link to external video on poster image in widget control preview so user can easily see the video referenced. See also #78 which would allow for the URL to be restored in the frame.

melchoyce:
> > It does seem like utilizing stills from the video providers might be a good way to solve this. We would still need to have some sort of fallback state where perhaps we show a generic video icon that is linked to the external site?
> 
> Yes. A generic video icon should be shown in that case.

If this is still the case, let's use https://developer.wordpress.org/resource/dashicons/#format-video.

westonruter:
Another option is to use [`wp_mime_type_icon()`](https://developer.wordpress.org/reference/functions/wp_mime_type_icon/).

westonruter:
Nevermind. That is just a rasterization of the Dashicon 😄 

https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/images/media/video.png

swissspidy:
> I think the reason for why post_ID is required is because currently WP will cache oEmbed responses in postmeta

That's correct. There's a ticket for that though: https://core.trac.wordpress.org/ticket/34115

> An oEmbed endpoint, via the REST API or via admin-ajax, would essentially be implementing the guts of `\WP_oEmbed::get_html()`, minus the `data2html()` call at the end

Indeed. We can extract that part into a new `\WP_oEmbed::get_data()` method that can be used by a proxy endpoint. I'm happy to help with that if needed. It's probably easiest to do this in a core ticket.

westonruter:
@swissspidy cheers, I've opened https://core.trac.wordpress.org/ticket/40450

timmyc:
Did a re-read of some of the concerns expressed at https://core.trac.wordpress.org/ticket/32417#comment:116 - the big one seems to be 

> Most default themes (and as a result, likely other themes) that implement custom styles for the MediaElement players scope their CSS to .hentry to have higher specificity than the MediaElement styling provided by WordPress.

The proposed solution of

> Add a .hentry class to the media widget container for all of them. Not semantic with the intent of that class and might risk causing other issues if themes use that class for other styling.

Seems easy enough, but I agree seems kind of hackish to do so.  I don't know what the standard procedures are in situations like this where a new core feature could potentially create a visual regression on existing themes.

I suppose my .02 on the matter is, this Video widget is entirely new - so not sure if it should be considered a blocker that themes in the wild have handled legacy usage of `mediaelement` players in a certain manner by targeting `.hentry`.  A fix would be simple by targeting `.widget widget_media_video` in a similar manner... but really seems like the root problem should be addressed in its own core fix.

timmyc:
In 2542ad9 I have added a placeholder/overlay for the poster images using the dashicon:

__YouTube__
customize__wordpress_develop_ _just_another_wordpress_site_

__Externally Hosted m4v file__
customize__wordpress_develop_ _just_another_wordpress_site_

westonruter:
> I suppose my .02 on the matter is, this Video widget is entirely new - so not sure if it should be considered a blocker that themes in the wild have handled legacy usage of mediaelement players in a certain manner by targeting .hentry. A fix would be simple by targeting .widget widget_media_video in a similar manner... but really seems like the root problem should be addressed in its own core fix.

I also think that adding a `.hentry` class would be a hack and it could even mess up Microformat parsers. To me it seems core themes should be updated to have the right styles if they don't already have them, and any themes out there which need updated styles should also incorporate them for full compatibility with 4.8.

melchoyce:
I want to explore the placeholder design a bit more — I'll post up some ideas tomorrow.

melchoyce:
So, after looking at this — I think we can keep the icon centered on black, if there's no poster image we can pull in. However, if there is a poster image, putting the icon over it is actually quite tricky to get right. White doesn't work on every preview; if the preview is light, the icon overlaying it can look broken. Putting it on top of a background started to get busy. I also think the icon makes it look like you can click on the preview to play.

Instead, I think that if a video has a poster image, we shouldn't show an icon at all — just the poster.

Another option for the no-poster preview, if we want to remove the icon entirely, is to overlay some (translatable) text that says "no preview available," or something similar.

melchoyce:
One weird thing I noticed: if I update the external video, it saves that new video, even if I don't save the Customizer.

westonruter:
> One weird thing I noticed: if I update the external video, it saves that new video, even if I don't save the Customizer.

If you are reloading the customizer then this is expected behavior. If you do a straight reloaded the changeset query parameter will persist and the previous customized state will be restored. There should probably be some notice that alerts the user of a restored session (beyond the enabled Save & Publish button: this would make sense to do as part of the work on drafting.

westonruter:
I'm working on getting this branch updated with the latest from `master`.

westonruter:
@melchoyce does the link type make any sense in the context of the video widget? Is there a use case to add not the media itself but just a link to the media or to the attachment page?

image

westonruter:
I found a case where perhaps it makes sense to allow for videos to be added as links instead of embeds. Maybe the user wants to embed a link off to an external site and include just a poster image. I tried adding accommodating that in ceecca8, but as I think about it again, they should just then use the image widget itself for this. If we want to get rid of the link type for video images, I have a couple commits for this: https://github.com/xwp/wp-core-media-widgets/compare/add/video-widget...add/video-widget-without-link-type

timmyc:
Couldn't a generic Text/HTML widget be used for embed code too?  Usually those embed codes include all the necessary bits like a poster image.

westonruter:
> Couldn't a generic Text/HTML widget be used for embed code too? Usually those embed codes include all the necessary bits like a poster image.

I'm talking about when you go to edit a post, and you click Add Media, and you click on a video to add. You can then see the Attachment Display Settings there show the link dropdown have 3 options: embed (the default), file, and post. If you do anything other than embed, then it just outputs a link element; no player.

When adding an external video URL, I sometimes see a “Link Text” field show up when the parse-embed request has problems (as it does). There may be scenarios where the link text is needed as a fallback, but I haven't been able to pinpoint when and where the field is always accessible.

westonruter:
I'm playing around with adding tracks to the video, and I see it is currently broken. I upload a VTT file and then when I try to add the track, I get an exception:

screen shot 2017-04-18 at 08 50 07

melchoyce:
> @melchoyce does the link type make any sense in the context of the video widget? Is there a use case to add not the media itself but just a link to the media or to the attachment page?

Cut it ✂️

westonruter (approved):

PR #57, opened 2017-04-07
Make medium the default image size instead of full

### LINKED ISSUE #55: Default image size to "medium"
Seems like it could be a better default size than full for the majority of themes.

westonruter:
Fixes #55.


obenland (approved):
Makes sense

PR #58, opened 2017-04-08
Refactor MediaFrame and Library extensions to not depend on closures

westonruter:
This PR eliminates `Library` and `MediaFrame` classes from being extended inside a control's `initialize` method, which effectively meant that these classes get extended for each widget instance instead of once for all. The reason why they were defined in the `initialize` method was for the sake of getting access to a control's configuration, but this PR changes that so that the relevant configuration is passed in when the classes are instantiated. This feels much cleaner and the right way to do it.

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -23,6 +23,142 @@ wp.mediaWidgets = ( function( $ ) {
 	component.modelConstructors = {};
 
 	/**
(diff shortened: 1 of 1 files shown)


obenland (approved):
This is where I'd love to have some qUnit tests that I could run to double-check things behave as before.

Took the widget through a smoke test and it looks good.

westonruter:
@obenland good call. I merged the `improve-media-class-extensions`  branch into `add/qunit` locally and the tests passed.

PR #59, opened 2017-04-10
Prepare plugin for life after core merge

westonruter:
Once we do the initial core merge, we should be able to continue working on the plugin which should override what is in core so that the plugin always provides the latest and greatest state of the features.

- [ ] Figure out how to manage upgrading/overriding the PHP classes.


obenland:
> Figure out how to manage upgrading/overriding the PHP classes.

We could define empty/abstract core widget classes if they don't exist and extend them with plugin-specific ones.

```php
if ( ! class_exists( 'WP_Widget_Gallery' ) ) :
    class WP_Widget_Gallery {
        public function __construct()
    }
endif;

class WP_Widget_Gallery_4_9 extends WP_Widget_Gallery {}
```

westonruter:
@obenland I found an alternative method/hack for facilitating this plugin to continue to be developed even once the media widgets have been merged into core. I found `wp_maybe_load_widgets()` and realized I could replace it with our own forked version to load an old `default-widgets.php` and then the widget PHP classes from our plugin instead of the ones that would have been included in core.

Naturally we can wait to merge this until we're about to commit the patch to `trunk`.

PR #65, opened 2017-04-10
Remove align from image widget instance props; hide alignment fields

### LINKED ISSUE #61: Remove floats
We included every attributed afforded to regular images, but floats don't really work in this context, and could potentially mess up people's layouts. Let's remove.

westonruter:
Fixes #61

### DIFF
--- a/core-media-widgets.php
+++ b/core-media-widgets.php
@@ -106,7 +106,7 @@ function wp32417_add_classname_to_display_settings() {
 			if ( ! templateEl ) {
 				return;
 			}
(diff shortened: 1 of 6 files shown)


westonruter:
In select frame:

image

In edit frame:

image

obenland:
Looks good! We should be able to remove the custom Twenty Sixteen styles too then.

obenland (approved):
I think we've made every reasonable effort here to prevent users from aligning their images. 

For completeness sake I should note that it's still possible for users to float the image (not the caption though) if they add an `align*` class to the "Image CSS Class" option. Which I'm fine with. At that point they should also be prepared for the results.

Issue #67, opened 2017-04-10
Media meda shouldn't sync automatically when selecting media in the media widget


adamsilverstein:
see https://core.trac.wordpress.org/ticket/40403

PR #68, opened 2017-04-10
Don't sync media meta

adamsilverstein:
Fixes https://github.com/xwp/wp-core-media-widgets/issues/67

### DIFF
--- a/wp-admin/js/widgets/media-image-widget.js
+++ b/wp-admin/js/widgets/media-image-widget.js
@@ -1,5 +1,5 @@
 /* eslint consistent-this: [ "error", "control" ] */
-(function( component ) {
+(function( component, $ ) {
 	'use strict';
 
 	var ImageWidgetModel, ImageWidgetControl;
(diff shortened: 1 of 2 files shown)


wp-admin/js/widgets/media-image-widget.js
```diff
@@ -191,8 +191,11 @@
 			mediaFrame.state( 'replace-image' ).on( 'replace', updateCallback );
 			mediaFrame.on( 'close', function() {
 				mediaFrame.detach();
+				wp.media.model.Attachment.prototype.sync = defaultSync;
 			});
 
+			var defaultSync = wp.media.model.Attachment.prototype.sync;
+			wp.media.model.Attachment.prototype.sync = function() { return $.Deferred().rejectWith( this ).promise(); };
```
westonruter: @adamsilverstein correct me if I am wrong, but I don't believe that the changes here in `editMedia` are not necessary because the `image-details` state doesn't attempt to sync any of the changes back to the server, right?

adamsilverstein: You are probably right, let me test that and update the PR.

westonruter:
@melchoyce do you concur that the attachment syncing behavior should be disabled for the media widget?

melchoyce:
Is that what [we talked about in Slack](https://wordpress.slack.com/archives/C0381N237/p1491846231444237)? If so, yes.

westonruter (approved):

PR #70, opened 2017-04-10
Make WP_Widget_Media::get_instance_schema() public

### LINKED ISSUE #51: Make WP_Widget_Media::get_instance_schema() public
Currently the `WP_Widget_Media::get_instance_schema()` method is `protected` for the sake of:

> This is protected because it may become part of WP_Widget eventually.

See [#35574](https://core.trac.wordpress.org/ticket/35574) (Add REST API JSON schema information to WP_Widget).

The idea was that it could be made `protected` while we wait for #35574 to land. Nevertheless, if that ticket results in `get_instance_schema` becoming `public` then any existing media widget extensions in plugins which retain `protected` visibility will then fatal error, which can be seen here: https://3v4l.org/j6emh

So I think we should just go ahead and make it `public` now, especially if we can validate that #35574 is indeed desired and should become part of `WP_Widget` itself.

westonruter:
Fixes #51.

### DIFF
--- a/core-media-widgets.php
+++ b/core-media-widgets.php
@@ -83,13 +83,13 @@ function wp32417_twentyten_styles() {
  */
 function wp32417_widgets_init() {
 	require_once( dirname( __FILE__ ) . '/wp-includes/widgets/class-wp-widget-media.php' );
-	/* TODO: require_once( dirname( __FILE__ ) . '/wp-includes/widgets/class-wp-widget-audio.php' ); */
-	require_once( dirname( __FILE__ ) . '/wp-includes/widgets/class-wp-widget-image.php' );
-	/* TODO: require_once( dirname( __FILE__ ) . '/wp-includes/widgets/class-wp-widget-video.php' ); */
+	/* TODO: require_once( dirname( __FILE__ ) . '/wp-includes/widgets/class-wp-widget-media-audio.php' ); */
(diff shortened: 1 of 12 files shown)


westonruter:
Humm, maybe we should rename `WP_Widget_Media::get_instance_schema()` to `WP_Widget_Media::get_item_schema()` for parity with `WP_REST_Controller::get_item_schema()`?

westonruter:
See question at https://core.trac.wordpress.org/ticket/35574#comment:24

westonruter:
I think that we should go ahead and define the `get_properties_schema()` method on `WP_Widget_Media` and go ahead and mark it as `public`. If we determine that [#35574](https://core.trac.wordpress.org/ticket/35574) should proceed, we can also define `WP_Widget::get_properties_schema()` on the base class. If there is by chance a change in the return value, then we can change the name in the base class. But this is a useful method in the context of the media widget, and it seems somewhat peculiar to limit access to this one method when there are other `public` methods also added to `WP_Widget_Media`.

See conversation at https://wordpress.slack.com/archives/C02RQC26G/p1491891894123842

obenland:
If we don't call it item schema, why do we need to change the naming? "Instance" is a familiar term used throughout all widgets

obenland (approved):
❤️

PR #73, opened 2017-04-12
Harden logic for rendering preview

### LINKED ISSUE #66: When image is deleted, only show 'Select Image' link
When I use an image for the widget from my media library and later delete it, the widget interface correctly shows that the image is missing:

The buttons below the image are incorrect, it should only show 'Select Image' as in the default state of the image when no image is selected:

westonruter:
Moves `error` property from `selectedAttachment` model onto `control.model`, as suggested by @obenland.

Fixes #66.

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -735,7 +735,6 @@ wp.mediaWidgets = ( function( $ ) {
 			var property = $( this ).data( 'property' );
 			attributes[ property ] = $( this ).val();
 		} );
-		delete attributes.id; // Read only.
 
 		// Suspend syncing model back to inputs when syncing from inputs to model, preventing infinite loop.
 		widgetControl.stopListening( widgetControl.model, 'change', widgetControl.syncModelToInputs );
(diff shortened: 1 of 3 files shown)


wp-admin/js/widgets/media-widgets.js
```diff
@@ -414,11 +399,15 @@ wp.mediaWidgets = ( function( $ ) {
 		/**
 		 * Whether a media item is selected.
 		 *
-		 * @return {boolean} Whether selected.
+		 * @return {boolean} Whether selected and no error.
 		 */
 		isSelected: function isSelected() {
 			var control = this;
 
+			if ( control.model.get( 'error' ) ) {
```
timmyc, line 407: I thought I had written some qunit tests around `isSelected` but they must have gotten lost in the shuffle on that PR.  This method, and the changes here, could use some test coverage.  I'd be happy to add that in here on this branch if you'd like.

obenland (approved):
Tests well!

westonruter:
@timmyc please open a new PR for the qunit test, thanks.

PR #76, opened 2017-04-12
Add more isSelected tests

timmyc:
Quick follow-up to #73, adding in more test cases to cover the various states of `isSelected`


westonruter (approved):
👍

PR #77, opened 2017-04-12
Hide selection pane when no attachment isSelected()

### LINKED ISSUE #75: Attachment details show before image is selected
screen shot 2017-04-12 at 11 37 18 am

You can see here that I don't have an image selected yet, but the attachment details are showing up anyway (empty)

timmyc:
Fixes #75.  In my testing, this approach seems to work well - but due to not fully groking all things `wp.media` yet, not certain if there are side-effects it could cause.

__Before__
Nexus5 Viewport:

Note when an empty `selection` is passed, the Add to Widget button is enabled here:

__After__
This looks much nicer on smaller viewports:

And large viewport, looks just like the add media dialog when first opened in the editor now, and the Add to Widget button is disabled since no media item is selected yet:

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -431,7 +431,7 @@ wp.mediaWidgets = ( function( $ ) {
 		selectMedia: function selectMedia() {
 			var control = this, selection, mediaFrame, defaultSync;
 
-			selection = new wp.media.model.Selection( [ control.selectedAttachment ] );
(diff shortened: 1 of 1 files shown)


westonruter:
@timmyc this solution is almost exactly what I was envisioning when I saw this issue reported.

westonruter (approved):
@timmyc the one issue I saw was when you click to Change Media and the user had previously supplied an image by URL instead of selecting an uploaded `attachment`. In that case, the initial selection also needs to be empty. See f04b331.

PR #83, opened 2017-04-13
Improve alignment JS code style with core standards

### LINKED ISSUE #80: JS coding standards
I'd recommend a general review before merging in core.

Minor things like spaces around arguments have patterns to follow clearly explained here:
https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#arrays-and-function-calls

Ideally, all these things (and more) should be addressed to avoid introducing even more inconsistencies in core.

Documentation:
https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/

While not explicitly mentioned, for analogy with PHP standards there should be a proper alignment, for example for the @param type, name, and description. Example from the current code:
```
* @param {wp.media.view.MediaFrame.Select} mediaFrame Select frame.
* @param {object} attachment Attachment object.
```

should be
```
* @param {wp.media.view.MediaFrame.Select} mediaFrame Select frame.
* @param {object}                          attachment Attachment object.
```

I'd also argue that `wp.media.view.MediaFrame.Select` is not a JS type :) I know this is common practice but IMHO should indicate just the proper JS type and `wp.media.view.MediaFrame.Select` should go in the description.

One more example:
```
* @param {Object}         options - Options.
* @param {Backbone.Model} options.model - Model.
* @param {jQuery}         options.el - Control container element.
```

here the alignment is a bit better, I wouldn't know what the `-` hyphen represents, and descriptions should be aligned. Same consideration for the type.

One more example, here the alignment looks good but I wouldn't know what the square brackets around param names are:
```
 * @param {string|Object} key       Attribute name or attribute pairs.
 * @param {mixed|Object}  [val]     Attribute value or options object.
 * @param {Object}        [options] Options when attribute name and value are passed separately.
```

westonruter:
Fixes #80.

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -87,7 +87,7 @@ wp.mediaWidgets = ( function( $ ) {
 		/**
 		 * Create the default states.
 		 *
-		 * @return {void}
+		 * @returns {void}
 		 */
 		createStates: function createStates() {
 			this.states.add( [
@@ -306,7 +306,7 @@ wp.mediaWidgets = ( function( $ ) {
 		/**
 		 * Update the selected attachment if necessary.
 		 *
-		 * @return {void}
+		 * @returns {void}
 		 */
 		updateSelectedAttachment: function updateSelectedAttachment() {
 			var control = this, attachment;
@@ -354,7 +354,7 @@ wp.mediaWidgets = ( function( $ ) {
 		/**
 		 * Get template.
 		 *
-		 * @return {Function} Template.
+		 * @returns {Function} Template.
 		 */
 		template: function template() {
 			var control = this;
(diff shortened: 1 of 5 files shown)


wp-admin/js/widgets/media-widgets.js
```diff
@@ -614,20 +614,20 @@ wp.mediaWidgets = ( function( $ ) {
...
 	/**
 	 * Collection of all widget model instances.
 	 *
 	 * @type {Backbone.Collection}
 	 */
-	component.modelCollection = new ( Backbone.Collection.extend( {
+	component.modelCollection = new ( Backbone.Collection.extend({
 		model: component.MediaWidgetModel
-	} ) )();
+	}) )();
```
timmyc, line 630: Should this be `}))()`?

timmyc (approved):
Wasn't sure about the one line, otherwise looks good to me... and I will make sure to not use spaces like this in future code 🤠

wp-admin/js/widgets/media-widgets.js
westonruter: Maybe, maybe not. The convention is to only collapse whitespace following the brace. There's no convention for this specific case other than the blanket:

> Use spaces liberally throughout your code. “When in doubt, space it out.”
>
> These rules encourage liberal spacing for improved developer readability.

So I think unless the standards explicitly say there should be no space, then there _should_ be a space.

PR #84, opened 2017-04-13
Improve handling of embedded media (via external URL)

### LINKED ISSUE #78: Pre-populate URL field with existing URL selected when opening embed frame
In the case where a user enters a URL to embed instead of selecting an attachment, it would probably also be desirable for the user if the URL they entered were pre-populated into the URL field for embedding:

image

We do this presently for attachments, passing the `selectedAttachment` as the `selection` to show in the attachments browser sidebar.

- [x] Also, if a `url` was selected, then the “Insert from URL” should be the initial frame not the browser..

See also #75.

westonruter:
* Open media modal with `embed` frame pre-selected if the media is embedded as opposed to referencing an attachment.
* Populate existing (image) properties in insert embed frame. This allows the URL to be modified and all of the existing properties to be retained.
* Hide alignment controls in embed frame; remove obsolete `align` reference.
* Add new `widget_id` attribute to Backbone model and define as `idAttribute` to prevent confusion with attachment `id` and `attachment_id`.
* Refactor logic for mapping model props with media frame props
* Define media frame prop in model instance schema.
* Eliminate need for subclassed image widget `getSelectFrameProps` with its private methods.
* Ensure that size change in edit frame pre-populates size in select frame.
* Prevent empty width and height from causing external image from having zero dimensions.

Fixes #78.


westonruter:
This is turning out very well!

Previously if you wanted to modify the properties of an embed, you were _required_ to click the “Edit Media” button:

screen shot 2017-04-13 at 17 17 46

Now, however, these same fields will also get populated into the media insertion frame when you click “Change Media”:

screen shot 2017-04-13 at 17 18 03

timmyc:
@westonruter is this still WIP or is it ready to be tested / reviewed?

westonruter:
@timmyc still WIP. I want to work out the mapping logic better.

westonruter:
@timmyc ok, now it's ready for review 😄 I hope that the changes here will reduce the amount of code and duplication that media subclasses will need, so it should reduce the amount of code needed in #53, though integrating the video changes here may illuminate shortcomings that weren't accounted for.

timmyc:
> I hope that the changes here will reduce the amount of code and duplication that media subclasses will need

🎉  I will give this a test today

wp-admin/js/widgets/media-widgets.js
```diff
@@ -429,21 +428,26 @@ wp.mediaWidgets = ( function( $ ) {
...
+			var control = this, selection, mediaFrame, defaultSync, mediaFrameProps;
 
 			if ( control.isSelected() && 0 !== control.model.get( 'attachment_id' ) ) {
 				selection = new wp.media.model.Selection( [ control.selectedAttachment ] );
 			} else {
 				selection = null;
 			}
 
+			mediaFrameProps = control.mapModelToMediaFrameProps( control.model.toJSON() );
+			control.displaySettings.set( 'size', mediaFrameProps.size );
```
timmyc: Isn't `size` image specific?  Could we just check the presence of `mediaFrameProps.size` prior to this line?

wp-admin/js/widgets/media-widgets.js
```diff
@@ -502,20 +506,86 @@ wp.mediaWidgets = ( function( $ ) {
...
+		getModelPropsFromMediaFrame: function getModelPropsFromMediaFrame( mediaFrame ) {
+			var control = this, state, mediaFrameProps;
+
+			state = mediaFrame.state();
+			if ( 'insert' === state.get( 'id' ) ) {
+				mediaFrameProps = _.extend(
+					state.get( 'selection' ).first().toJSON(),
+					mediaFrame.content.get( '.attachments-browser' ).sidebar.get( 'display' ).model.toJSON()
+				);
+				if ( mediaFrameProps.sizes[ mediaFrameProps.size ] ) {
```
timmyc: Could `mediaFrameProps.sizes` be null here on any non-image instances?

wp-admin/js/widgets/media-widgets.js
```diff
@@ -502,20 +506,86 @@ wp.mediaWidgets = ( function( $ ) {
...
+					mediaFrameProps.url = mediaFrameProps.sizes[ mediaFrameProps.size ].url;
+				}
+			} else if ( 'embed' === state.get( 'id' ) ) {
+				mediaFrameProps = _.extend(
+					state.props.toJSON(),
+					{
```
timmyc: Could this object be dynamically generated on a per-subclass basis?  Like for a video not having the `size` attribute.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -502,20 +506,86 @@ wp.mediaWidgets = ( function( $ ) {
...
+			return modelProps;
+		},
+
+		/**
+		 * Map model props to media frame props.
+		 *
+		 * @param {Object} modelProps - Model props.
+		 * @returns {Object} Media frame props.
+		 */
+		mapModelToMediaFrameProps: function mapModelToMediaFrameProps( modelProps ) {
```
timmyc, line 574: Nice 👍

wp-admin/js/widgets/media-widgets.js
westonruter: Good question. It is image specific, but the underlying display settings model seems to populate regardless: https://github.com/WordPress/wordpress-develop/blob/4.7.3/src/wp-includes/js/media-views.js#L1201

Even when it's not an image: https://github.com/WordPress/wordpress-develop/blob/4.7.3/src/wp-includes/js/media-views.js#L1231-L1241

But humm, the template does seem to check for existence: https://github.com/WordPress/wordpress-develop/blob/4.7.3/src/wp-includes/media-template.php#L680-L708

So yeah, better safe than sorry and do an explicit check here. 👍

wp-admin/js/widgets/media-widgets.js
westonruter: Yes, that would probably make it more intuitive when reading the source. Ultimately unrecognized properties will get dropped when the props get passed through `mapMediaToModelProps`, but definitely the "full" value here is definitely image-specific and it should not be here in the base class.

westonruter: OK, see 7aa50495.

timmyc (approved):
Great updates here @westonruter - testing out well for me too.

PR #85, opened 2017-04-14
Remove flex layout on buttons

### LINKED ISSUE #49: Image widget buttons and flexbox
A few issues quickly discussed on Slack about buttons and flexbox, maybe something to consider:
- the image widget buttons use flexbox; while the Customizer has dropped support for old IEs, the widgets screen hasn't
- it would be the first time flexbox is used in core (by core I mean not the customizer), maybe this should be discussed with the project leads, namely Helen :slightly_smiling_face:
- longer translations of the buttons may break everything, see screenshot below (`flex-wrap: wrap` plus some adjustments could help here but I'm not sure that would be the desired effect)

screen shot 2017-04-05 at 23 18 00

See also https://core.trac.wordpress.org/ticket/40220

timmyc:
This branch fixes #49, and aligns with the proposed natural-width buttons in core ( https://core.trac.wordpress.org/ticket/40220 )

__Before__
before

__After__
after

__Long Strings__
long-string

### DIFF
--- a/wp-admin/css/widgets/media-widgets.css
+++ b/wp-admin/css/widgets/media-widgets.css
@@ -14,7 +14,7 @@
 }
 
 .media-widget-buttons .button {
-	margin: 10px 0 0 10px;
+	margin: 8px 0 0 8px;
 	width: auto;
 }
 .media-widget-control:not(.selected) .media-widget-buttons .button,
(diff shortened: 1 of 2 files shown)


melchoyce:
Awesome :) 

Can we left-align the buttons?

westonruter:
@melchoyce like so?

image

westonruter (approved):

timmyc:
@westonruter looks like you had a beauty day at Hoodoo.  That view of Mt. Washington is divine from up there in the winter <3

PR #86, opened 2017-04-14
Flesh out eslint and jscs rules for jsdoc

westonruter:
I wanted to also add the `requireDescriptionCompleteSentence` and `enforceExistence` rules for JSCS, but they caused more problems than they solved.

It also seems like the `require-jsdoc` ESLint rule isn't applying to our class methods.

See #80.


timmyc (approved):
Looks good.

PR #89, opened 2017-04-19
Fix image_title and title issues

timmyc:
It appears #84 may have introduced some regressions into the image widget around the `title` and `image_title` attributes.  The first issue can be seen where the widget `title` is deleted upon selection of media ( which do not have an `image_title` assigned to them ):

title-lost

The other time this collision between the widget `title` and the `image_title` happens can be seen when the mediaFrame image title is not persisted, and always shows the widget `title`:

title-collisions

The fix proposed here is to skip over the widget `title` attribute in `mapModelToMediaFrameProps` and adding in the proper `media_prop` for `image_title`.


westonruter (approved):
Good catches.

PR #90, opened 2017-04-20
Re-use oEmbed response to obtain thumbnail_id and set as poster image

westonruter:
Here's an experiment to piggyback on the existing oEmbed request that we're making to obtain the preview to then also populate the `poster`.

Amends #53.

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -87,17 +87,20 @@ wp.mediaWidgets = ( function( $ ) {
 		/**
 		 * Refresh embed view.
 		 *
-		 * Forked override of {wp.media.view.Embed#refresh()} to suppress irrelevant "link text" field.
+		 * Forked override of {wp.media.view.Embed#refresh()} to suppress irrelevant "link text" field
+		 * and so that the thumbnail_id for oEmbeds can be captured for use as the poster frame.
 		 *
 		 * @returns {void}
 		 */
 		refresh: function refresh() {
-			var type = this.model.get( 'type' ), Constructor;
+			var type = this.model.get( 'type' ), PatchedConstructor, Constructor;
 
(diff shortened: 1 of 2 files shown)


westonruter:
One thing that gives me pause about this is the setting of the `poster` image and how the poster image is showing in the browser:

image

Maybe this isn't actually a problem with this PR at all but exposes a problem with assigning posters in general. The `.mejs-poster` element in `mediaelementplayer.min.css` has `background-size:contain` but as can be seen above this doesn't look right. Changing the style to 
`background-size:cover` looks much better:

image

This may also be related to not having the `height`/`width` being considered.

Nevertheless, if I disable MediaElement.js entirely and let the browser natively render the video player (via plugin `add_filter( 'wp_video_shortcode_library', '__return_empty_string' );`), the poster shows (in Chrome) with effectively `background-size:contain` though the `width`/`height` are very wrong:

image

wp-admin/js/widgets/media-widgets.js
```diff
@@ -87,17 +87,20 @@ wp.mediaWidgets = ( function( $ ) {
...
 		refresh: function refresh() {
-			var type = this.model.get( 'type' ), Constructor;
+			var type = this.model.get( 'type' ), PatchedConstructor, Constructor;
 
 			if ( 'image' === type ) {
 				Constructor = wp.media.view.EmbedImage;
 			} else if ( 'link' === type ) {
-				Constructor = wp.media.view.EmbedLink.extend( {
+
+				// This should be eliminated once #40450 lands of when this is merged into core.
```
timmyc, line 102: s/of/or here ` lands or when`

timmyc (approved):
I agree with your concerns on the poster image - in my testing of a few youtube videos it really looks pretty rough when viewed in the rendered widget.  I'm fine with getting this merged into the main branch and iterating further.

PR #91, opened 2017-04-20
Fix Preview Image Flash

### LINKED ISSUE #74: Attached image preview flash
ezgif-2-dc989101e0

When previewing an attachment image in the widget preview and the selected size is not the full size, there is a flash of the selected image before the full attachment object has loaded and the image is replaced with its full size.

Should we wait showing a preview of the image until the attachment object has loaded?
Should the image size in the preview be a medium, like the default widget image size, rather than full?

timmyc:
This branch seeks to close #74 - currently the image preview shows the full size image even when a "thumbnail" is selected.  The proposed fix here is to more closely mimic the logic that is used on the front-end widget display and set the image `src` to the appropriate URL within the preview as well.  

~A nice side-effect of this is the image flash no longer occurs.  I opted to create a stand-alone method for `getImagePreviewUrl` so we could have some tests around it.~

Fixes #74.


wp-admin/js/widgets/media-image-widget.js
```diff
@@ -25,17 +25,41 @@
...
+		getImagePreviewUrl: function getImagePreviewUrl() {
+			var control = this, attachmentId, size, attachmentSizes;
+			attachmentId = control.model.get( 'attachment_id' );
+			size = control.model.get( 'size' );
+
+			// Attempt to return the specific attachment image size requested.
+			attachmentSizes = control.selectedAttachment.get( 'sizes' );
+			if ( attachmentId && 'custom' !== size && attachmentSizes && attachmentSizes[ size ] ) {
+				return attachmentSizes[ size ].url;
+			}
```
westonruter: @timmyc I think in practice this will always end up being the same as the `control.model.get( 'url' )` since the URL for the specific size chosen: https://github.com/xwp/wp-core-media-widgets/blob/0502cd6d7a04fa5bd74fb21bae56a3a3cb38dcaf/wp-admin/js/widgets/media-widgets.js#L516-L518

So I wonder if we need this function and an just replace calls to `control.getImagePreviewUrl()` with simply `control.model.get( 'url' )`?

timmyc: Yeah I thought the same thing but in my testing I was running into scenarios when the image url was not respecting the size chosen.  Let me take it for another spin and see if I can reproduce/track down the root cause if that is the case.

timmyc: It was the template change that was the actual fix 😆  - sometimes I just have a splendid knack for overcomplicating things.

westonruter (approved):
Originally I was thinking the widget preview should show a more abstract normalized representation of the selected image, such as the `medium` size. I also thought that if it was a huge image then it would be a waste to show it in the small space. But the image is going to be shown in the preview anyway 😄 and the dimensions of the image in the control preview will likely be similar to the image in the frontend sidebar, so it doesn't really matter.

Issue #54, opened 2017-04-07
Refactor MediaWidgetControl#renderPreview() to return rendered template 


westonruter:
Currently the JS that a media subclass writes for the `renderPreview` method looks something like this:

```js
var control = this, previewContainer, previewTemplate;
previewContainer = control.$el.find( '.media-widget-preview .rendered' );
previewTemplate = wp.template( 'wp-media-widget-video-preview' );
previewContainer.html( previewTemplate( { attachment: control.selectedAttachment.attributes } ) );
```

The logic that a subclassed method needs to implement could be simplified down to just:

```js
var previewTemplate = wp.template( 'wp-media-widget-video-preview' );
return previewTemplate( { attachment: this.selectedAttachment.toJSON() } );
```

There can be another method in `MediaWidgetControl` which is responsible for calling `renderPreview` and then injects the markup returned to inject into the proper DOM element.

westonruter:
I guess `maybelater`.

PR #92, opened 2017-04-20
Only renderPreview when certain props change

timmyc:
This branch seeks to fix one of the TODO items in #53 

> Prevent re-rendering video preview when none of the fields have changed, e.g. when changing the title.

I had noticed this in prior PR reviews where the image preview was being re-rendered upon each key-up on the title as well.  This wasn't optimal for the image widget, but the video widget the experience is noticeably rough when a media player is in the render... or when we are issuing a CORS request to retrieve the poster image for externally hosted files.

The proposed fix here is to introduce a new value to the field schema called `should_preview_update`.  So for fields that _shouldn't_ trigger a `renderPreview`, we set the value to `false`.  The logic in `shouldPreviewUpdate` then evaluates the model changes against this field to determine if we should indeed re-render the preview.

For the `ImageWidgetControl` I have set `shouldPreviewUpdate` to `false` for all fields that are not used in the preview.  So the only attributes that should trigger a re-render are `url`, `alt`, `id/attachment_id`, and `size`/`width`/`height`.

Best way to test this is to inspect the preview image on a widget, and modify the title... note that the DOM shouldn't update while you change the title.  Likewise, changing the caption in the media modal should not re-render the preview... but changing the size, or the image should.


westonruter:
@timmyc good work. I was thinking about taking a similar approach of defining a subset of the model props that are relevant to updating the preview. What I wasn't sure about was where to define these. Storing them in the schema as `should_preview_update` works well, though maybe `is_preview_template_prop` maybe feels a bit better.

An idea for how we could streamline the template rendering, and perhaps eliminate `shouldPreviewUpdate` and `maybeRenderPreview` could be to introduce a new `previewTemplateProps` model which we sync changes from the subset of `model` as defined by `should_preview_update` into this new `previewTemplateProps` model, and then we can just continue to re-render the preview whenever any of the `previewTemplateProps` model changes. Do you think this would simplify things?

timmyc:
> could be to introduce a new previewTemplateProps model which we sync changes from the subset of model

I think this could totally work too - since we are already doing the `_.merge` of `selectedAttachment` and `model` - might be cleaner to track the pertinent changes in a stand-alone model like `previewTemplateProps`.

I forget which PR we had chatted about this before, but this change could also lead to simplifying `renderPreview` - perhaps we could just store the templateID on the view, and the base `MediaWidgetControl` could `renderPreview` for all widget types.

westonruter:
> I forget which PR we had chatted about this before, but this change could also lead to simplifying renderPreview - perhaps we could just store the templateID on the view, and the base MediaWidgetControl could renderPreview for all widget types.

Is it #54? I was almost going to close it out due to how #53 had async behavior in `renderPreview`. But then #90 made it synchronous again so it would now be possible to have one base `renderPreview` method that is called whenever `previewTemplateProps` changes, if the template ID is defined by a property on the subclasses as you suggest. The main concern of the subclasses then would be able to amend the `previewTemplateProps` with additional dynamic properties beyond what is defined in the `model`, such as the `imageSrc` and `imageFilename`.

I suppose `renderPreview` could also be made to accommodate asynchronicity if we define it to return a promise which resolves with the template string instead of returning the template string directly.

timmyc:
@westonruter in c0c443d I implemented a the new `previewTemplateProps` model.  Tests have been updated as well.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -122,6 +122,7 @@ public function get_instance_schema() {
 				'default' => '',
 				'sanitize_callback' => 'sanitize_text_field',
 				'description' => __( 'Title for the widget' ),
+				'should_preview_update' => false,
```
westonruter, line 125: If we eventually move to using React or Vue, then this `should_preview_update` would be irrelevant, correct?

timmyc: Correct this is PR is very much like having a `shouldComponentUpdate`, or a connected component that only listens to certain changes in the state tree and re-renders accordingly.

Between here and there though, I still think we need this logic to prevent re-rendering the preview on each `keyup` in the title.  Especially if we are hitting API endpoints or oembed providers.

timmyc:
@westonruter outside of a rebase is there any other changes you would like to see here?

westonruter (approved):
I think this looks good.

PR #96, opened 2017-04-21
[WIP] Fix video dimensions

### LINKED ISSUE #93: Handle height/width for videos
There is no `height` or `width` fields when selecting or editing a video in the media manager. This is a challenge. The `video` shortcode defaults the width to `640` and the height as `360`. When MediaElement.js initializes, it resizes the video to fit inside of its container scaled, but if MediaElement.js is disabled to let the browser natively render the video player (via plugin `add_filter( 'wp_video_shortcode_library', '__return_empty_string' );`), the widget overflows the container:

image

If a `max-width:100%` is hacked onto the video container element, and the `height` attribute is removed as follows:

```diff
diff --git a/wp-includes/widgets/class-wp-widget-media-video.php b/wp-includes/widgets/class-wp-widget-media-video.php
index f772817..dfc82f7 100644
--- a/wp-includes/widgets/class-wp-widget-media-video.php
+++ b/wp-includes/widgets/class-wp-widget-media-video.php
@@ -135,6 +135,8 @@ class WP_Widget_Media_Video extends WP_Widget_Media {
 			return;
 		}
 
+		add_filter( 'wp_video_shortcode', array( $this, 'inject_video_max_width_style' ) );
+
 		echo wp_video_shortcode(
 			array_merge(
 				$instance,
@@ -142,6 +144,20 @@ class WP_Widget_Media_Video extends WP_Widget_Media {
 			),
 			$instance['content']
 		);
+
+		remove_filter( 'wp_video_shortcode', array( $this, 'inject_video_max_width_style' ) );
+	}
+
+	/**
+	 * Inject max-width and remove height.
+	 *
+	 * @param string $html Video shortcode HTML output.
+	 * @returns string HTML Output.
+	 */
+	public function inject_video_max_width_style( $html ) {
+		$html = preg_replace( '/\bheight="\d+"/', '', $html );
+		$html = str_replace( 'style="', 'style="max-width:100%;', $html );
+		return $html;
 	}
 
 	/**
```

The result is better for the native `video` element: 

image

However, when MediaElement.js is re-enabled then the video displayed shorter than the native player:

image

Upon playing, however, it then resizes dynamically to have the proper height for the video.

westonruter:
- [ ] Reset width/height for non-oEmbeds.
- [ ] Actually fix the issue with sizing in #93.

Fixes #93.

### DIFF
--- a/tests/phpunit/test-class-wp-widget-media-video.php
+++ b/tests/phpunit/test-class-wp-widget-media-video.php
@@ -33,6 +33,8 @@ class Test_WP_Widget_Media_Video extends WP_UnitTestCase {
 					'title',
 					'url',
 					'content',
+					'width',
+					'height',
 				),
 				wp_get_video_extensions()
 			),
@@ -172,6 +174,30 @@ class Test_WP_Widget_Media_Video extends WP_UnitTestCase {
 			'h4x' => 'value',
 		), $instance );
 		$this->assertSame( $result, $instance );
+
+		// Width and height are handled.
+		$result = $widget->update(
+			array(
+				'width' => -1,
+				'height' => 'bad',
+			),
+			array()
+		);
+		$this->assertEmpty( $result );
+		$result = $widget->update(
+			array(
+				'width' => 3,
+				'height' => '4',
+			),
+			array()
+		);
+		$this->assertSame(
+			$result,
+			array(
+				'width' => 3,
+				'height' => 4,
+			)
+		);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


westonruter:
Instead of adding support for video width and height, should the fixed width and height be removed altogether from the rendered video output?

While not ideal to have `!important` properties, including the following CSS in the page could work:

```css
.widget_media_video .wp-video {
    width: auto !important;
    height: auto !important;
}
.widget_media_video video {
    display: block;
    height: auto;
    max-width: 100%;
}
```

I'm not sure if this is better than doing regex on the inline styles (per diff seen in #93) and how this relates to the `object-fit` CSS property.

I wonder why the `video` shortcode doesn't do something responsive like this by default.

@kopepasah thoughts?

PR #97, opened 2017-04-24
Add visual text widget

westonruter:
See: 
https://core.trac.wordpress.org/ticket/35243
https://core.trac.wordpress.org/ticket/35760

# Todo
- [x] Figure out which `the_content` filters we can apply on the content. If not `the_content` itself, consider adding common `the_content` filters to the `widget_text` filter.
- [x] Add a filter specifically for the visual text widget for plugins to add their own `the_content` filters.
- [x] Ensure that TinyMCE editor gets rebuilt upon saving widget on admin screen.
- [x] Fix appearance of button tooltips in editor.
- [x] Fix appearance of link toolbar.
- [x] Re-build editor when widget is dragged to a new position in the DOM.
- [ ] ~Fix selective refresh to use fallback full refresh when content includes `script` tag.~ (Requires core selective refresh patch, see [comment](https://github.com/xwp/wp-core-media-widgets/pull/97#issuecomment-299077176) and core ticket [#40658](https://core.trac.wordpress.org/ticket/40658).)
- [ ] ~Attempt to prevent TinyMCE from having to be rebuilt when saving widget on widgets admin screen, where a flicker could happen. (Consider listening for clicks on the Save widget button to detach the TinyMCE editor, and then restore upon `widget-updated`.)~
- [x] Testing across supported browsers


mehigh:
I'll take this opportunity to familiarise myself with the js widgets.

melchoyce:
Still a WIP? UI looks good, but I can't type into it yet.

Did we decide if this is replacing the existing text widget, or is it in addition to the existing text widget? Will both ever live on the same site?

westonruter:
@melchoyce You can't type into it? I'm able to:

image

You do have to be on `trunk`.

And in talking with @azaozz the idea was that it should be an enhancement on top of the existing Text widget. It wouldn't be a new widget type.

melchoyce:
Nope, can't seem to type in it. I can type into the text tab, and if I click on the formatting buttons, the formatting tags appear in the text tab, but nothing shows up in the visual tab. Here's a quick video: https://cloudup.com/cDpQDHR6VqW

I'm on `4.8-alpha-39357-src`.

> And in talking with @azaozz the idea was that it should be an enhancement on top of the existing Text widget. It wouldn't be a new widget type.

Great. Will the name go back to just being "text" when it's merged? The widget name should stay the same.

westonruter:
> Great. Will the name go back to just being "text" when it's merged? The widget name should stay the same.

Yes. I renamed it “Visual Text” for this PR to just make it easer to see that the plugin was working. I can rename it back now since you can just look at a widget now.

westonruter:
> Nope, can't seem to type in it. I can type into the text tab, and if I click on the formatting buttons, the formatting tags appear in the text tab, but nothing shows up in the visual tab. Here's a quick video: https://cloudup.com/cDpQDHR6VqW

I don't have any idea what is happening on your environment? 😦 

@obenland Are you able to replicate it?

westonruter:
Nevermind, I see the issue. It seems that when I add a _new_ widget it cannot be edited but any existing Text widgets I previously added can be edited.

obenland:
I'm not able to replicate it either, wfm. Are there any errors in the console @melchoyce?

westonruter:
@melchoyce please try again. For some reason it seems the editor sometimes doesn't fully initialize due to a race condition. I added a delay in 4b583ab to work around it, but we should figure out why exact the issue is happening. It may be due to the container not being displayed at the time of calling.

wp-admin/js/widgets/text-widgets.js
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		widgetControl = new component.TextWidgetControl({
+			el: widgetContent
+		});
+
+		// Add delay as without this the iframe may not get contenteditable.
+		_.delay( function() {
+			widgetControl.render();
+		}, 100 ); // eslint-disable-line no-magic-numbers
```
westonruter: @azaozz Any suggestions for a way to get around this issue with `wp.editor.initialize()` not successfully adding `contenteditable` for the editor iframe when the `textarea` was not visible at the time of invocation?

westonruter: The `fast` animation for `slideDown` is `200` milliseconds, but having `100` here seems to work. If the delay here is `10` it doesn't seem to be enough.

melchoyce:
Works great now!

mehigh:
According to my tests the wp.editor.initialize function call takes ±100ms to run - both on first initialisation, and on subsequent TinyMCE initialisations just as well. 
This seems to match the delay you experienced here with it requiring 100ms of 'visibility' before someone can interact with them. 

Correlation != causality... but my guess is that this is the browser's way to protect the computer resources by delaying any scripting that belongs to iframe's that are not visible.
I would recommend hiding it with something else than display: none (maybe position: absolute; left: -999em; instead) ... and see whether the browser behaves differently and allows the scripting to take place even before the user interacts with it and the iframe turns into becoming visible.

wp-admin/js/widgets/text-widgets.js
mehigh: According to my tests the wp.editor.initialize function call takes ±100ms to run - both on first initialisation, and on subsequent TinyMCE initialisations just as well.
This seems to match the delay you experienced here with it requiring 100ms of 'visibility' before someone can interact with them.

Correlation != causality... but my guess is that this is the browser's way to protect the computer resources by delaying any scripting that belongs to iframe's that are not visible.
I would recommend hiding it with something else than display: none (maybe position: absolute; left: -999em; instead) ... and see whether the browser behaves differently and allows the scripting to take place even before the user interacts with it and the iframe turns into becoming visible.

westonruter: @mehigh thanks for the info. If initialize takes ~100ms to run then the additional delay of 100ms makes sense, since `fast` animation for `slideDown` is 200ms. 

I'm not sure of the feasibility of changing how the `textarea` is hidden since it is inside of a `display:none` container that gets slid-down, like is done for all widgets.

wp-includes/widgets/class-wp-widget-visual-text.php
```diff
@@ -0,0 +1,160 @@
...
+		$instance = $old_instance;
+		$instance['title'] = sanitize_text_field( $new_instance['title'] );
+		if ( current_user_can( 'unfiltered_html' ) ) {
+			$instance['text'] = $new_instance['text'];
+		} else {
+			$instance['text'] = wp_kses_post( $new_instance['text'] );
+		}
+
+		// Eliminate filter from here on.
+		unset( $instance['filter'] );
```
westonruter: Maybe this is not a good idea, for example it could cause problems if a WP install downgrades from 4.8 to 4.7. Instead of removing `filter` maybe we should instead force `filter` to always be `true` and then to _add_ a new property, like `visual` which also gets set to `true`. Then when calling `widget` to render an instance, we can check for the presence of `visual` instead of `! isset( $instance['filter'] )`. Then the `filter` instance property would just persist as a back-compat property during 4.8 and could potentially be removed in the future.

westonruter: Updated in fbaef1b.

wp-admin/js/widgets/text-widgets.js
```diff
@@ -0,0 +1,197 @@
...
+	 * Handle widget being added or initialized for the first time at the widget-added event.
+	 *
+	 * @param {jQuery.Event} event - Event.
+	 * @param {jQuery}       widgetContainer - Widget container element.
+	 * @returns {void}
+	 */
+	component.handleWidgetAdded = function handleWidgetAdded( event, widgetContainer ) {
+		var widgetContent, widgetForm, idBase, widgetControl, widgetId, animatedCheckDelay = 50, widgetInside, renderWhenAnimationDone;
+		widgetForm = widgetContainer.find( '> .widget-inside > .form, > .widget-inside > form' ); // Note: '.form' appears in the customizer, whereas 'form' on the widgets admin screen.
+		widgetContent = widgetForm.find( '> .widget-content' );
```
timmyc: Could this be moved below the `return` on line 94 - so the logic isn't evaluated if not needed?

wp-admin/js/widgets/text-widgets.js
```diff
@@ -0,0 +1,197 @@
...
+	/**
+	 * Handle widget being added or initialized for the first time at the widget-added event.
+	 *
+	 * @param {jQuery.Event} event - Event.
+	 * @param {jQuery}       widgetContainer - Widget container element.
+	 * @returns {void}
+	 */
+	component.handleWidgetAdded = function handleWidgetAdded( event, widgetContainer ) {
+		var widgetContent, widgetForm, idBase, widgetControl, widgetId, animatedCheckDelay = 50, widgetInside, renderWhenAnimationDone;
+		widgetForm = widgetContainer.find( '> .widget-inside > .form, > .widget-inside > form' ); // Note: '.form' appears in the customizer, whereas 'form' on the widgets admin screen.
```
timmyc, line 153: This logic along with determining `idBase` is repeated in `handleWidgetUpdated` too.  Also any performance concerns with looking up these elements in the DOM frequently?

wp-admin/js/widgets/text-widgets.js
```diff
@@ -0,0 +1,197 @@
...
+	 *
+	 * This gets called via the 'widget-updated' event when saving a widget from
+	 * the widgets admin screen and also via the 'widget-synced' event when making
+	 * a change to a widget in the customizer.
+	 *
+	 * @param {jQuery.Event} event - Event.
+	 * @param {jQuery}       widgetContainer - Widget container element.
+	 * @returns {void}
+	 */
+	component.handleWidgetUpdated = function handleWidgetUpdated( event, widgetContainer ) {
```
timmyc, line 202: So I was trying to figure out the flow of things here a bit and added some logging statements to this method - and I was never able to trigger this logic.  In an older version of this branch, it seemed this logic was called on any `change` event in MCE but for now I can't seem to make it happen.

wp-includes/widgets/class-wp-widget-visual-text.php
```diff
@@ -0,0 +1,166 @@
...
+				 */
+				$text = apply_filters( 'widget_text_content', $widget_text, $instance, $this );
+
+			} elseif ( $instance['filter'] ) {
+				$text = wpautop( $text ); // Back-compat for instances prior to 4.8.
+			}
+		}
+
+		echo $args['before_widget'];
+		if ( ! empty( $title ) ) {
```
timmyc, line 83: In the other widgets we return if the main content is empty ( i.e. an image is missing ) - should we not render anything including the `$title` if `$text` is empty?

timmyc (commented):
A few questions below, also found some issues while testing out in Chrome:

- When expanding/collapsing the widget, there is still some oddness on the animation that is causing the MCE area to resize during the animation
- It appears the `link` toolbar button does not function in both visual and text modes.  Guessing to dialog might be hidden behind something else in the customizer view - though I could not find it.
- Tooltips on the toolbar buttons are not visible.  Wondering if this might be an accessibility concern.
- Changing the location of the widget in the customizer via drag-n-drop makes the Visual tab in MCE 👻  away

mce-goes-away

wp-admin/js/widgets/text-widgets.js
westonruter: The `handleWidgetAdded` function will only be called once for every widget created. So there is no performance concerns here.

wp-admin/js/widgets/text-widgets.js
westonruter: Yes

wp-admin/js/widgets/text-widgets.js
westonruter: It should never actually get triggered in the customizer, unless the PHP on the server generates new `input` elements which cause the widget to not be able to align the old fields with the new fields (when alignment succeeds, the customizer triggers `widget-synced` event instead; when alignment fails, an Apply button is shown after which clicking a `widget-updated` event will be triggered). On the widgets admin screen, however, there is none of this field syncing. Instead, you hit the Save button for a widget and then when the server returns the `widget-updated` event triggers.

wp-includes/widgets/class-wp-widget-visual-text.php
westonruter: No, because the `WP_Widget_Text` in core doesn't short-circuit, and so we should retain that old behavior for back-compat.

timmyc:
Looks like the visual editor going awry after moving the widget might be related to https://core.trac.wordpress.org/ticket/35243#comment:27

```
Note that the editor, once initialized, cannot be moved in the DOM. This is a security limitations for iframes in all browsers. If moving is necessary, it has to be unloaded, then initialized again.
```
So looks like we might need to listen for the remove and re-init MCE again after the move completes...

westonruter:
@timmyc great feedback:

> When expanding/collapsing the widget, there is still some oddness on the animation that is causing the MCE area to resize during the animation

Is it specifically with the horizontal resizing, where the widget inside container isn't full-width? Could it be this: https://core.trac.wordpress.org/ticket/31476#comment:31

> It appears the link toolbar button does not function in both visual and text modes. Guessing to dialog might be hidden behind something else in the customizer view - though I could not find it.

Seems like a bug in `wp.editor.initialize()`. @azaozz Thoughts?

> Tooltips on the toolbar buttons are not visible. Wondering if this might be an accessibility concern.

This could also be a problem with `wp.editor.initialize`. @azaozz Thoughts?

> Changing the location of the widget in the customizer via drag-n-drop makes the Visual tab in MCE 👻 away

Ouch. Yeah, that's a problem. We'll need to essentially invoke the `widget-updated` logic to re-build the editor after such re-ordering events.

timmyc:
Noting here that the issues with tooltips on buttons, and link dialog not showing are due to z-index conflicts.  The customizer overlay, `div.wp-full-overlay` has a `z-index: 500000` whereas the mce values are all in the `130000` range

westonruter:
That reminds me, we encountered this `z-index` issues in Customize Posts as well. The workaround there: https://github.com/xwp/wp-customize-posts/blob/538d6f82b5d3de5a4fa1102743a82c5ef92de3ea/css/customize-posts.css#L223-L240

timmyc:
z-index fix in 01caa27 ... dark dark times when stuff like that is written.

mehigh:
Thanks @timmyc for the fix. I needed to do the same when using the TinyMCE editor in a WP admin overlay. Can't those old inline styles just disappear already?

westonruter:
> When expanding/collapsing the widget, there is still some oddness on the animation that is causing the MCE area to resize during the animation

@timmyc is this still an issue after https://core.trac.wordpress.org/changeset/40569 ?

westonruter:
Something like this core patch will ensure scripts in selective refresh responses get executed properly:

```diff
--- src/wp-includes/js/customize-selective-refresh.js
+++ src/wp-includes/js/customize-selective-refresh.js
@@ -415,6 +415,11 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
 				return false;
 			}
 
+			if ( -1 !== placement.addedContent.indexOf( '<script' ) ) {
+				partial.fallback( new Error( 'script_content' ), [ placement ] );
+				return false;
+			}
+
 			/* jshint ignore:start */
 			self.orginalDocumentWrite = document.write;
 			document.write = function() {
```

westonruter:
Actually, the core patch for selective refresh should look like the following. We need to strip out IE conditional comments before we check if there are script tags, as otherwise there are full-page refreshes for any changes to video or audio widgets, since they have HTML5 shims:

```diff
--- src/wp-includes/js/customize-selective-refresh.js
+++ src/wp-includes/js/customize-selective-refresh.js
@@ -398,7 +398,7 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
 		 * @returns {boolean} Whether the rendering was successful and the fallback was not invoked.
 		 */
 		renderContent: function( placement ) {
-			var partial = this, content, newContainerElement;
+			var partial = this, content, newContainerElement, contentWithoutConditionalComments;
 			if ( ! placement.container ) {
 				partial.fallback( new Error( 'no_container' ), [ placement ] );
 				return false;
@@ -415,6 +415,12 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
 				return false;
 			}
 
+			contentWithoutConditionalComments = placement.addedContent.replace( /<!--\[if.+?\]>(.|\n)+<!\[endif\]-->/g, '' );
+			if ( /<script/.test( contentWithoutConditionalComments ) ) {
+				partial.fallback( new Error( 'script_content' ), [ placement ] );
+				return false;
+			}
+
 			/* jshint ignore:start */
 			self.orginalDocumentWrite = document.write;
 			document.write = function() {
```

westonruter:
Please test core patch: https://core.trac.wordpress.org/ticket/40658

westonruter:
Regarding:
> Attempt to prevent TinyMCE from having to be rebuilt when saving widget on widgets admin screen, where a flicker could happen. (Consider listening for clicks on the Save widget button to detach the TinyMCE editor, and then restore upon `widget-updated`.)

I think rebuilding the editor is just a reality of TinyMCE living in a widget. We have to rebuild the editor when dragging a widget around in the DOM. Adding the same workaround done for media widgets to embed the widget controls outside of the container that gets replaced would have back-compat concerns for other plugins that extend the Text widget. Eventually with a next-generation of JS-powered widgets, we can avoid the full HTML replacement of the form every time the widget is saved on the widgets admin screen. So I think we should leave this as-is for now.

wp-admin/js/widgets/text-widgets.js
```diff
@@ -0,0 +1,259 @@
...
+				editor.on( 'blur', function() {
+					editor.off( 'change', onChanged );
+					triggerChangeIfDirty();
+				} );
+			}
+
+			/**
+			 * Watch an iframe for the destruction of its TinyMCE contenteditable contents.
+			 *
+			 * @todo There may be a better way to listen for an iframe being destroyed.
```
westonruter, line 119: Any idea for a better way to listen for the destruction of an iframe's content document?

westonruter:
AFAIK there are no further known bugs for this feature.

obenland:
@timmyc Can this be merged?

timmyc:
Just tested in IE11, Firefox 53 - all seems to be working as expected.  🚢

timmyc (approved):

PR #105, opened 2017-04-26
Remove/hide the media replacement UI

### LINKED ISSUE #100: Replace button in Edit Image window duplicates functionality of Change Image action
The `Replace` button in the `Edit Image`/`Image Details` modal duplicates the functionality of the `Change Image` button.

I'd suggest removal of the button from `Edit Image`, unless there's a different UX reason for leaving it in.

screen shot 2017-04-24 at 1 52 55 pm
screen shot 2017-04-24 at 1 53 02 pm

### LINKED ISSUE #102: Replace Button Clickable When No Item Selected
When editing an existing media widget ( Audio, Image, or Video ) you have the option to 'Replace' the existing asset.  The ability to replace in this flow is under question in #100 - but there is also a bug in the replace flow.

When using the 'Replace' functionality in any of the widgets, the Replace button is always active, even when a media item *is not* selected.  Clicking on the Replace button without an item selected results in error being thrown:

__Video__
customize__blog_ _wordpress_develop_and_plugins_ _fly_fishers_place_ _wordpress_com_and_wp-core-media-widgets_ _timmyc_lilninja_ _-zsh_ _178x43

__Audio__
customize__blog_ _wordpress_develop_and_plugins_ _fly_fishers_place_ _wordpress_com_and_wp-core-media-widgets_ _timmyc_lilninja_ _-zsh_ _178x43

__Image__

It appears these issues might actually be in underlying core media files... but at the very least, the Replace button should not be enabled unless a new media item is selected.

westonruter:
Fixes #102
Fixes #100


timmyc:
Should I create a trac issue for #102 as it still is broken in core media flows?

wp-admin/js/widgets/media-audio-widget.js
```diff
@@ -82,11 +113,12 @@
 			metadata = control.mapModelToMediaFrameProps( control.model.toJSON() );
 
 			// Set up the media frame.
-			mediaFrame = wp.media({
+			mediaFrame = new AudioDetailsMediaFrame({
```
westonruter, line 116: Having to extend the entire class to remove the `replace-audio` seems like it should be unnecessary. It seems like all we really should have to do is:

```js
mediaFrame.states.remove( 'replace-audio' );
```

But it's not working. What also doesn't work is:

```js
mediaFrame.toolbar.view.states.remove( 'replace-audio' )
``` 

Nor does this work:

```js
mediaFrame.states.get( 'replace-audio' ).deactivate()
```

wp-admin/css/widgets/media-widgets.css
```diff
@@ -47,7 +47,8 @@
 .media-frame.media-widget .image-details .embed-media-settings .setting.align,
 .media-frame.media-widget .attachment-display-settings .setting.align,
 .media-frame.media-widget .embed-media-settings .setting.align,
-.media-frame.media-widget .embed-link-settings .setting.link-text {
+.media-frame.media-widget .embed-link-settings .setting.link-text,
+.media-frame.media-widget .replace-attachment {
```
westonruter, line 51: There is no underlying state that controls whether or not the Replace button is shown. So we have to rely on CSS instead.

westonruter:
> Should I create a trac issue for #102 as it still is broken in core media flows?

@timmyc yes.

wp-admin/js/widgets/media-audio-widget.js
timmyc: Ugh that is too bad the other approaches don't work.  It would be nice if we could also pass an `option` for `states` to the `wp.media` constructor too.

timmyc (approved):
I agree it isn't optimal to have to `extend` to replace the states, but at least it is quite explicit reading the code what all is going on.

This is testing out well for me.

Issue #36, opened 2017-03-20
Test image widget on supported browsers and devices


westonruter:
- [x] Full support back to IE11, partial back to IE8.
- [x] Firefox, Chrome, Safari: Current – 1.
- [x] iOS Safari: Current – 1.
- [x] Android browser: Current major.
- [x] Chrome/Android: Current.

Add comments below with relevant screenshots and which browsers/devices have been tested. Once done, check of the boxes.

timmyc:
IE8 Widget View:
![ie8_-_win7__running_](https://cloud.githubusercontent.com/assets/22080/24878946/f4901d72-1de9-11e7-83cd-bd0a917c3605.png)

Note "Select Image" button and "No Image selected" placeholder are both shown even when an image is selected.

Otherwise working as expected in IE8 with media library selections and adding via URL.

timmyc:
@westonruter when you say "partial back to IE8" what exactly does that mean?  Should the issue of the Select Image and placeholder above be fixed?

westonruter:
@timmyc “partial back to IE8” means it's functional but the styling may not be quite right.

It looks like the problem here is IE8's lack of support for `:not()` 😦 

So yeah, that needs to be fixed. This would be a good thing to fix while also fixing #66 and #61.

timmyc:
Okay IE9 Results, I'm hitting a JS error with anything customizer related in general.  Wondering if someone else had a VM handy to test but, the following is triggered when simply adjusting the title on any widget ( even non-media widgets ):

![ie9_-_win7__running_](https://cloud.githubusercontent.com/assets/22080/24881448/a39ae402-1df2-11e7-872d-8772464d527a.png)

This also prevents any edits from happening within the medial modal as well - nothing saves once this exception occurs.

This is the line that is throwing the error: https://core.trac.wordpress.org/browser/trunk/src/wp-admin/js/customize-controls.js#L4780

Edits working as expected from the Widget manager.

timmyc:
Perhaps the error is related to https://core.trac.wordpress.org/changeset/39686/trunk/src/wp-admin/js/customize-controls.js

westonruter:
@timmyc youch! Good catch! I can see that the change there did not sufficiently check for IE9 compat. I think a change like this is needed:

```diff
--- src/wp-admin/js/customize-controls.js
+++ src/wp-admin/js/customize-controls.js
@@ -4762,6 +4762,12 @@
 			 */
[...]

PR #110, opened 2017-04-26
Maintain button margin on small screens

obenland:
There is a default mobile style for customizer buttons that is more specific than our selector was.

See https://github.com/xwp/wp-core-media-widgets/issues/36#issuecomment-296788423

Before:
screen shot 2017-04-24 at 11 49 21 am

After:
screen shot 2017-04-26 at 9 04 20 am


timmyc (approved):
LGTM and testing out well.

timmyc:
Looks like after a rebase this is good to go @obenland

PR #112, opened 2017-04-26
Don't pass a selection or metadata on selectMedia

### LINKED ISSUE #103: URL set when clicking "Change Image"
When clicking "Change Image" for an existing image that is set via URL - the current image widget URL is set in the selection window.  Performing the same flow when changing an image from the media library, the current image is not selected:

__Change Image: Current Image set via URL__

__Change Image: Current Image from media library__

Note in the media library selection, the attachment details are shown in this selection view too, and they probably should not.

While not a blocker, if one is trying to change images via URL you have to select all/remove the existing URL to enter a new one.

timmyc:
Fixes #103 - This proposed change is to always have the `selectMedia()` method open the media modal in `select` state, regardless of if a media item is currently set in the widget.  To me the intention of clicking "Change Media" in the widget is to do just that - select an entirely new media item.  Currently when clicking "Change Media" the existing media item details are populated in the modal which I feel users could find confusing.

Curious for your thoughts on this idea @westonruter - or if there is another reason for keeping the `embed` state active when clicking "Change Media"

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -572,29 +572,22 @@ wp.mediaWidgets = ( function( $ ) {
 		 * @returns {void}
 		 */
 		selectMedia: function selectMedia() {
-			var control = this, selection, mediaFrame, defaultSync, mediaFrameProps;
+			var control = this, selection, mediaFrame, defaultSync;
 
 			if ( control.isSelected() && 0 !== control.model.get( 'attachment_id' ) ) {
 				selection = new wp.media.model.Selection( [ control.selectedAttachment ] );
-			} else {
-				selection = null;
-			}
-
-			mediaFrameProps = control.mapModelToMediaFrameProps( control.model.toJSON() );
-			if ( mediaFrameProps.size ) {
-				control.displaySettings.set( 'size', mediaFrameProps.size );
 			}
 
 			mediaFrame = new component.MediaFrameSelect({
 				title: control.l10n.select_media,
 				frame: 'post',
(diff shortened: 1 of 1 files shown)


westonruter:
> Curious for your thoughts on this idea @westonruter - or if there is another reason for keeping the embed state active when clicking "Change Media"

The idea was that if you supplied an external URL then if you wanted to tweak the URL, for example change the query params or fix a typo you made, you could do so by modifying the URL that is re-populated into that field. Maybe this isn't a common-enough use case to warrant this.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -603,11 +596,7 @@ wp.mediaWidgets = ( function( $ ) {
 				var attachment = {}, state = mediaFrame.state();
 
 				// Update cached attachment object to avoid having to re-fetch. This also triggers re-rendering of preview.
-				if ( 'embed' === state.get( 'id' ) ) {
-					_.extend( attachment, { id: 0 }, state.props.toJSON() );
-				} else {
-					_.extend( attachment, state.get( 'selection' ).first().toJSON() );
-				}
+				_.extend( attachment, state.get( 'selection' ).first().toJSON() );
```
westonruter: The `embed` state is needed for when the user inserts an external URL, as opposed to selecting from the media library.

timmyc: Doh, adding it back in.

westonruter:
Closing per https://github.com/xwp/wp-core-media-widgets/issues/103#issuecomment-298436011

PR #113, opened 2017-04-27
Utilize wp_attachment_is() in addition to strtok(post_mime_type) comparison

westonruter:
See https://github.com/xwp/wp-core-media-widgets/commit/030ead1fc52b0461a0dfb6d740756bc0c272431e#commitcomment-21926912


wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -147,7 +147,7 @@ public function is_attachment_with_mime_type( $attachment, $mime_type ) {
 		if ( 'attachment' !== $attachment->post_type ) {
 			return false;
 		}
-		return strtok( $attachment->post_mime_type, '/' ) === $mime_type;
+		return wp_attachment_is( $mime_type, $attachment ) || strtok( $attachment->post_mime_type, '/' ) === $mime_type;
```
obenland: Do you expect situations where file extensions are not covered by `wp_attachment_is()`?

westonruter: I can't think of a situation right now. Fine to restrict to `wp_attachment_is()` for now.

obenland (approved):
LGTM 👍

PR #115, opened 2017-04-28
Add ability to reset props to defaults when media changes

### LINKED ISSUE #106: Poster Image not Cleared on Video Change
When changing a video from an external url ( i.e. YouTube ) and selecting a video file from the media library instead, the old poster image from the YouTube video is persisted and still shown.

__Steps to Recreate__
- Add a new video widget
- Set the URL to a YouTube video
- Confirm the poster image is shown in the `renderPreview`
- Click "Change Video"
- Select a video file from the media library
- Note in the `renderPreview` the old poster image is shown from the YouTube video, but if you click play, the video file from the library is properly set.

timmyc:
This branch seeks to fix #106 by adding a new schema attribute `reset_on_media_change` that the callback in `selectMedia()` utilizes to determine if any `model` attributes should be reset back to their default value.

The initial use-case for this is clearing out of the `poster` image when changing videos from a YouTube video, back to a locally hosted video - but there may be other attributes that we determine should be reset back to the default when the media changes.

Here I trigger the logic of media change when the `url` attribute changes, but I almost feel like we should just key off of `attachment_id` instead.... but using `url` handles the case where you are changing between externally hosted videos, and the second one chosen not having a poster image returned in the oembed response  ( edgiest of edge cases there ).

__To Test__
- Add a new video widget
- Set the URL to a YouTube video
- Confirm the poster image is shown in the `renderPreview`
- Click "Change Video"
- Select a video file from the media library
- Note in the `renderPreview` that the old YouTube poster image is not shown.


westonruter:
If the poster is removed then #116 then this PR won't have any props to reset.

timmyc:
Indeed with #116 in, I don't think we need this anymore.

PR #116, opened 2017-04-30
Remove video poster

### LINKED ISSUE #95: Improve rendering of video poster images
Improve rendering of `poster` images when ME.js will show YouTube or Vimeo players with their own poster frames underneath.

westonruter:
Closes #95.


obenland:
Should other `poster` references in in `tests/qunit/index.html` be removed as well? Or do we continue to need those?

westonruter:
@obenland those references are coming from underlying media templates from core, so they're not relevant for removal.

timmyc:
With the removal of poster support, is #115 even needed?

wp-includes/widgets/class-wp-widget-media-video.php
```diff
@@ -59,16 +59,10 @@ public function get_instance_schema() {
...
-					'type' => 'string',
-					'default' => '',
-					'format' => 'uri',
-					'description' => __( 'URL to the poster frame' ),
-				),
 				'preload' => array(
 					'type' => 'string',
 					'enum' => array( 'none', 'auto', 'metadata' ),
-					'default' => 'none',
+					'default' => 'auto',
```
timmyc: Was just curious about this change to `auto` is that the typical default in the editor video flow?

timmyc (commented):
I will give this a more thorough test in a bit, but just had the one question above.

wp-includes/widgets/class-wp-widget-media-video.php
westonruter: Good catch. The default for video should actually be `metadata`, per https://github.com/WordPress/wordpress-develop/blob/9e07667864ca860d8683e30ccfdca796229c5657/src/wp-includes/media.php#L2463

timmyc (approved):
Tested this out, and indeed it is working as expected and takes care of the stale poster image when changing videos.

Worth noting that without the poster image, external videos show up like the following in the preview:

/cc @melchoyce as I know you were considering changing this design a bit.

melchoyce:
Are we not able to grab a still from the video itself if it's external?

melchoyce:
Here's my ideal scenario:
- Your video should always show a preview image
- You should never have to manually select a preview image inside of WordPress

I thought removing the poster image option would make this the case, but from your screenshot @timmyc I think I was mistaken?

westonruter:
@timmyc maybe we need to look at resurrecting the async logic for rendering the preview so that it can fetch the `thumbnail_url` via the oEmbed proxy (or the `noembed.com` polyfill)? Or rather, maybe we can continue with synchronous rendering but then asynchronously set the `img.src` once the oEmbed request returns?

timmyc:
>  maybe we need to look at resurrecting the async logic for rendering the preview so that it can fetch the thumbnail_url via the oEmbed proxy (or the noembed.com polyfill)? Or rather, maybe we can continue with synchronous rendering but then asynchronously set the img.src once the oEmbed request returns?

Both sound reasonable to me.  We can try those out in a follow-up PR too.

westonruter:
OK, let's follow-up in a subsequent PR then.

PR #117, opened 2017-05-01
Defer adding hooks until the widgets are registered

### LINKED ISSUE #98: Ensure widget is still registered at running hook
Since hooks are added in the constructor (not ideal) when the hook runs it should ideally check to make sure the widget is still enqueued.

westonruter:
Fixes #98.

### DIFF
--- a/wp-includes/widgets/class-wp-widget-media.php
+++ b/wp-includes/widgets/class-wp-widget-media.php
@@ -83,7 +83,7 @@ abstract class WP_Widget_Media extends WP_Widget {
 	/**
 	 * Add hooks while registering all widget instances of this widget class.
 	 *
-	 * @since 2.8.0
+	 * @since 4.8.0
 	 * @access public
 	 */
 	public function _register() {
--- a/wp-includes/widgets/class-wp-widget-media.php
+++ b/wp-includes/widgets/class-wp-widget-media.php
@@ -78,6 +78,15 @@ abstract class WP_Widget_Media extends WP_Widget {
 			$widget_opts,
 			$control_opts
 		);
+	}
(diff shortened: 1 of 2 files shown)


obenland:
So you're saying let the unregistering party deal with cleanup?
(Which I'm absolutely fine with, just wanted to clarify)

westonruter:
@obenland no, this approach which you suggested will avoid the need to cleanup. if `unregister_widget()` is called before widgets are registered, as it would be, then the `_register` method would never get called on the unregistered widget class and there would be nothing to clean up.

obenland:
Oh, right, core widgets don't get registered until priority 100.

westonruter:
To clarify, as I see there is some overloading of the term “register”: when you call `register_widget()` it doesn't actually call `WP_Widget::_register()`. The `register_widget()` call happens at `widgets_init` priority 10, and then at `widgets_init` priority 100 when `WP_Widget_Factory::_register_widgets()` is called, and this factory method then actually iterates over all of the registered widgets (classes) and then registers all of the widget instances.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -78,6 +78,15 @@ public function __construct( $id_base, $name, $widget_options = array(), $contro
 			$widget_opts,
 			$control_opts
 		);
+	}
+
+	/**
+	 * Add hooks while registering all widget instances of this widget class.
+	 *
+	 * @since 2.8.0
```
obenland: Docs typo

obenland (approved):
One docs typo, looks good otherwise

PR #118, opened 2017-05-02
Populate link_url for non-custom linkTo settings

### LINKED ISSUE #111: Image Widget: Edit screen doesn't pick up file and post links
When images are linked to their file or attachment page, the url to that page never gets passed to the image model, resulting in the edit screen always assuming it's not linked at all.

To reproduce:
1. Select an image
2. For "Link to", either select "Media File" or "Attachment Page".
3. Click "Add to Widget"
4. Click "Edit Image"
5. Check the "Link to" field.

`mediaFrameProps` never get `linkUrl` set when `link` is either "file" or "post".

obenland:
This will properly populate the "Link To" setting in the image widget's edit modal.

The amount of code and its complexity is a poor representation of the time and effort that went into figuring out how to fix this bug. Thanks @westonruter for the guidance.

Fixes #111

### DIFF
--- a/wp-admin/js/widgets/media-widgets.js
+++ b/wp-admin/js/widgets/media-widgets.js
@@ -726,6 +726,12 @@ wp.mediaWidgets = ( function( $ ) {
 				modelProps.height = mediaFrameProps.customHeight;
 			}
 
+			if ( 'post' === mediaFrameProps.link ) {
+				modelProps.link_url = control.selectedAttachment.get( 'link' );
+			} else if ( 'file' === mediaFrameProps.link ) {
+				modelProps.link_url = mediaFrameProps.url;
+			}
+
 			// Because some media frames use `id` instead of `attachment_id`.
 			if ( ! mediaFrameProps.attachment_id && mediaFrameProps.id ) {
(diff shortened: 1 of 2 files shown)


wp-admin/js/widgets/media-widgets.js
```diff
@@ -726,6 +726,12 @@ wp.mediaWidgets = ( function( $ ) {
 				modelProps.height = mediaFrameProps.customHeight;
 			}
 
+			if ( 'post' === mediaFrameProps.link ) {
+				modelProps.link_url = control.selectedAttachment.get( 'link' );
```
westonruter: This feels a bit “impure” in that we're looking at `control.selectedAttachment` but we're specifically being passed `mediaFrameProps`. I don't suppose `mediaFrameProps.linkUrl` is possible?

obenland: Good point. Does 49e9e69 look better?

westonruter: Beautiful

westonruter (approved):

PR #119, opened 2017-05-02
Prevent Widget Title Loss

timmyc:
Quite similar to #89 - I encountered another issue where the video widget title was being over-written by attachment titles.

__Steps To Recreate Bug on master__
- Add a new video widget
- Change the title of the widget
- Select a movie file from the library
- Note the widget title is not the same as it was before

Repeat these same steps with this branch to ensure all is fixed.


westonruter (commented):
Good move. This is something I noticed but didn't feel right.

PR #120, opened 2017-05-02
Gallery Widget

### LINKED ISSUE #62: Implement gallery widget
## Gallery Widget Flow

gallery widget flow

See https://core.trac.wordpress.org/ticket/41914

obenland:
Fixes #62.

See https://core.trac.wordpress.org/ticket/41914

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -7,7 +7,7 @@
 
 	<rule ref="WordPress">
 		<exclude name="WordPress.XSS.EscapeOutput"/>
-		<exclude name="WordPress.Arrays.ArrayIndentationSniff"/>
+		<exclude name="WordPress.Arrays.ArrayIndentation"/>
 	</rule>
 
 	<rule ref="WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents">
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -7,6 +7,7 @@
 
 	<rule ref="WordPress">
 		<exclude name="WordPress.XSS.EscapeOutput"/>
+		<exclude name="WordPress.Arrays.ArrayIndentationSniff"/>
 	</rule>
 
 	<rule ref="WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents">
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -7,8 +7,6 @@
 
 	<rule ref="WordPress">
 		<exclude name="WordPress.XSS.EscapeOutput"/>
-		<exclude name="WordPress.Arrays.ArrayIndentation"/>
-		<exclude name="WordPress.Arrays.ArrayDeclarationSpacing"/>
 	</rule>
 
 	<rule ref="WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents">
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -8,6 +8,7 @@
 	<rule ref="WordPress">
 		<exclude name="WordPress.XSS.EscapeOutput"/>
 		<exclude name="WordPress.Arrays.ArrayIndentation"/>
+		<exclude name="WordPress.Arrays.ArrayDeclarationSpacing"/>
 	</rule>
 
 	<rule ref="WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents">
--- a/tests/qunit/test-media-image-widget.js
+++ b/tests/qunit/test-media-image-widget.js
@@ -31,7 +31,9 @@
 		equal( templateProps.currentFilename, 'wp-header-logo.png', 'mapModelToPreviewTemplateProps should set currentFilename based off of url' );
 		equal( templateProps.url, testImageUrl, 'mapModelToPreviewTemplateProps should return the proper url' );
 		equal( templateProps.alt, 'some alt text', 'mapModelToPreviewTemplateProps should return the proper alt text' );
-		equal( templateProps.link_type, undefined, 'mapModelToPreviewTemplateProps should ignore attributes that are not needed in the preview' );
+
+		// TODO Fix this upstream.
+		//equal( templateProps.link_type, 'none', 'mapModelToPreviewTemplateProps should ignore attributes that are not needed in the preview' );
 		equal( templateProps.error, false, 'mapModelToPreviewTemplateProps should return error state' );
 
 		// Test mapModelToPreviewTemplateProps() when error is set on model.
--- a/wp-admin/js/widgets/media-gallery-widget.js
+++ b/wp-admin/js/widgets/media-gallery-widget.js
@@ -92,7 +92,9 @@
 		editMedia: function editMedia() {
 			var control = this, selection, mediaFrame, defaultSync, mediaFrameProps;
 			if ( control.isSelected() && 0 !== control.model.get( 'selection' ) ) {
-				selection = new wp.media.model.Selection( JSON.parse( control.model.get( 'attachments' ) ) );
+				selection = new wp.media.model.Selection( JSON.parse( control.model.get( 'attachments' ) ), {
+					multiple: true
+				});
 			} else {
 				selection = null;
 			}
@@ -102,14 +104,16 @@
 				control.displaySettings.set( 'size', mediaFrameProps.size );
 			}
 			mediaFrame = new GalleryDetailsMediaFrame({
-				frame: 'select',
+				frame: 'manage',
 				text: control.l10n.add_to_widget,
 				selection: selection,
 				mimeType: control.mime_type,
 				selectedDisplaySettings: control.displaySettings,
 				showDisplaySettings: control.showDisplaySettings,
 				metadata: mediaFrameProps,
-				state: 'gallery'
+				editing:   true,
+				multiple:  true,
+				state: 'gallery-edit'
 			});
 			wp.media.frame = mediaFrame; // See wp.media().
 
@@ -118,6 +122,7 @@
 				var state = mediaFrame.state(), selectedImages;
 
 				selectedImages = selections || state.get( 'selection' );
+
 				if ( ! selectedImages ) {
 					return;
 				}
--- a/wp-admin/js/widgets/media-gallery-widget.js
+++ b/wp-admin/js/widgets/media-gallery-widget.js
@@ -53,25 +53,7 @@
 	 * @class GalleryWidgetModel
 	 * @constructor
 	 */
-	GalleryWidgetModel = component.MediaWidgetModel.extend( {
-		/**
-		 * Remove an attachment ID from attachments.
-		 *
-		 * @param {Integer} id - Attachment id to remove from attachments.
-		 * @returns {void}
-		 */
-		removeAttachmentId: function removeAttachmentId( id ) {
-			var attachments, newAttachments;
-			attachments = JSON.parse( this.get( 'attachments' ) );
-			newAttachments = _.filter( attachments, function( attachment ) {
-				return attachment.id !== id;
-			} );
-			this.set( {
-				'attachments': JSON.stringify( newAttachments ),
-				'ids': _.map( newAttachments, 'id' )
-			} );
-		}
-	} );
+	GalleryWidgetModel = component.MediaWidgetModel.extend( {} );
 
 	/**
 	 * Gallery widget control.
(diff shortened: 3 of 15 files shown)


timmyc:
Looks like @westonruter already commented but I saw this too:

`<br />↵<b>Notice</b>:  Array to string conversion in <b>/srv/www/wordpress-develop/public_html/src/wp-content/plugins/wp-core-media-widgets/wp-includes/widgets/class-wp-widget-media.php</b> on line <b>330</b><br />↵Array`

Which was resulting in some JS errors in the preview render.  I added some logic to ensure we are working against an array in the preview, but agree we should maybe consider returning just the image urls needed in attachments?  Not entirely sure the best thing to do.

🤞 that 905edbc also clears up the travis issues.

timmyc:
I [disabled the code sniffer](https://github.com/xwp/wp-core-media-widgets/pull/120/files#diff-b0d97627036a40d42d13ccff11b22a18R110) in a few lines as it was returning the following error about `rand`:

`wp-includes/widgets/class-wp-widget-media-gallery.php:112:38: error - Detected forbidden query_var "orderby" of "rand". Use vip_get_random_posts() instead. (WordPress.VIP.OrderByRand.orderby_orderby)`

Figured that didn't really apply here.

westonruter:
> I disabled the code sniffer in a few lines as it was returning the following error about rand

@timmyc feel free to give the `WordPress.VIP.OrderByRand.orderby_orderby` error a severity of `0` in the underlying `phpcs.xml.dist` instead.

westonruter:
Cross-reference Gallery block: https://github.com/WordPress/gutenberg/pull/740

timmyc:
@m1tk00 - excellent, I am not getting errors in the preview anymore!!  Outside of the linting issues we are having on the branch now - the following flow appears to be broken for me:

- Create a new Gallery Widget, assign photos, finish the modal wizard
- Click "Edit Gallery" in the widget.  Note the current photos are not selected, and the modal state shown is the same as creating a new gallery.

In this scenario, I expected to be shown the same dialog state as seen when clicking the 'pencil' to edit a gallery within the editor:

add_new_post_ _wordpress_develop_ _wordpress

m1tk00:
@timmyc yeah thats where I got stuck. I tried passing the data via the selected option but can't get it working

timmyc:
@m1tk00 so looking at how the Gutenberg Gallery block performs the operation: https://github.com/WordPress/gutenberg/blob/master/blocks/media-upload-button/index.js#L124-L141 - they hook into the `onOpen` event then add in the selected items at that point.  Do you want to give that a try or do you want me to work on that here?

m1tk00:
Ill try it in the morning and let you know :)

m1tk00:
@timmyc Ive managed to fix the edit button. It was calling a wrong media frame. Just pushed to the repo.

timmyc:
@m1tk00 I'm still having issues with the edit flow - my selections are not being shown when clicking the Edit button.

timmyc:
So the past 3 commits are me trying to get CI to pass :) - each time, there is an array test that seems to go :boom: like the latest:

`Parse error: syntax error, unexpected T_STRING in /tmp/wpcs/WordPress/Sniffs/Arrays/CommaAfterArrayItemSniff.php on line 10`

Any ideas @westonruter?  I tried to re-create the issue locally and could not, then again my phpcs knowledge is limited.

m1tk00:
@timmyc my bad, the file was not committed properly.

timmyc:
Adding in some notes here from some DM's with @m1tk00.  

- I am going to do a closer review of the code today and look for any issues
- @m1tk00 is going to work on getting some phpunit test coverage for the new widget
- I might try to add in some qunit tests as well
- @melchoyce I think we are nearing a point where we could use a little design direction on the widget.  As-is, the preview might look a bit _odd_ if a gallery has more than 3 images. @m1tk00 and I both thought of doing something similar to how Facebook handles large galleries to handle the small preview space gracefully 

facebook

wp-admin/js/widgets/media-gallery-widget.js
```diff
@@ -0,0 +1,254 @@
...
+				return $.Deferred().rejectWith( this ).promise();
+			};
+			mediaFrame.on( 'close', function onClose() {
+				wp.media.model.Attachment.prototype.sync = defaultSync;
+			});
+
+			mediaFrame.$el.addClass( 'media-widget' );
+			mediaFrame.open();
+
+			// Clear the selected attachment when it is deleted in the media select frame.
```
timmyc: Not sure we need this logic here... but will test to confirm

wp-includes/widgets/class-wp-widget-media-gallery.php
```diff
@@ -0,0 +1,231 @@
...
+	 * @access public
+	 *
+	 * @see WP_REST_Controller::get_item_schema()
+	 * @see WP_REST_Controller::get_additional_fields()
+	 * @link https://core.trac.wordpress.org/ticket/35574
+	 * @return array Schema for properties.
+	 */
+	public function get_instance_schema() {
+		return array_merge(
+			parent::get_instance_schema(),
```
timmyc: `get_instance_schema()` here will return `title`, `attachment_id`, and `url`.  We do need the title, but not `attachment_id` nor `url`.  Should we even call the parent method here, or just simply add in the `title` bits directly?

wp-includes/widgets/class-wp-widget-media-gallery.php
```diff
@@ -0,0 +1,231 @@
...
+	 *
+	 * @since 4.8.0
+	 * @access public
+	 */
+	public function render_control_template_scripts() {
+		parent::render_control_template_scripts();
+		?>
+		<script type="text/html" id="tmpl-wp-media-widget-gallery-preview">
+			<# var describedById = 'describedBy-' + String( Math.random() ); #>
+			<# data.attachments = data.attachments ? JSON.parse(data.attachments) : ''; #>
```
timmyc: Not sure if we need to be concerned or not, but should we verify that `data.attachments` is an Array here?

wp-includes/widgets/class-wp-widget-media-gallery.php
```diff
@@ -0,0 +1,231 @@
...
+			<# if ( data.error && 'missing_attachment' === data.error ) { #>
```
timmyc: Looks like we never set the `missing_attachment` error in this widget so we could likely remove this logic.

wp-includes/widgets/class-wp-widget-media-gallery.php
```diff
@@ -0,0 +1,231 @@
...
+				<div class="notice notice-error notice-alt notice-missing-attachment">
+					<p><?php echo $this->l10n['missing_attachment']; ?></p>
+				</div>
+
+			<# } else if ( Array.isArray( data.attachments ) && data.attachments.length ) { #>
+				<div class="gallery gallery-columns-{{ data.columns }}">
+					<# _.each( data.attachments, function( attachment, index ) { #>
```
timmyc: As mentioned in the https://github.com/xwp/wp-core-media-widgets/pull/120#issuecomment-323149823 I posted, we will likely want to revisit this preview logic and maybe only show the first 2-3 images in a larger gallery.

wp-includes/widgets/class-wp-widget-media-gallery.php
```diff
@@ -0,0 +1,231 @@
...
+	 * @param array $instance Widget instance props.
+	 * @return bool Whether widget has content.
+	 */
+	protected function has_content( $instance ) {
+
+		if ( ! empty( $instance['ids'] ) ) {
+
+			$attachments = explode( ',', $instance['ids'] );
+			foreach ( $attachments as $attachment ) {
+				if ( 'attachment' !== get_post_type( $attachment ) ) {
```
timmyc, line 233: Thinking out loud here, at which points in the code do we need to detect images that have been deleted from the media library, but are still attached to a gallery widget?

phpcs.xml.dist
```diff
@@ -7,6 +7,8 @@
 
 	<rule ref="WordPress">
 		<exclude name="WordPress.XSS.EscapeOutput"/>
+		<exclude name="WordPress.Arrays.ArrayIndentation"/>
```
timmyc: I need to remove these two lines now that @westonruter has fixed up the CI problems.

timmyc (commented):
Some notes and wonderings...

timmyc:
Okay took things for more of a test-drive and hit some issues along the way.  When creating a new gallery, the following error was thrown in the console, and the preview never renders:

I was able to edit an existing gallery - which is great! I did hit an issue though that it appears these options are not properly persisted from the media modal to the widget instance:

Likewise because the columns attribute doesn't seem to be persisted, things get a bit out of hand in the preview on larger galleries:

Oddly enough - in the gallery from the image above, I deleted the "troll king" image from the media library, but yet it was still showing in the preview, even after a hard refresh 🤔  - but on the front-end the gallery it is not shown:

gal_ _wordpress_develop

melchoyce:
I didn't encounter any console errors, but I did notice my gallery column settings didn't update in the preview when I changed them.

We'll also want to start figuring out how to render the gallery inside the widget as mocked up: https://github.com/xwp/wp-core-media-widgets/issues/62#issuecomment-294054977

melchoyce:
Also noticed the gallery images link to attachment pages by default. I noticed this might also be the default core behavior, which I think is kind of wonky, so I might submit a Trac ticket to change it.

timmyc:
Thanks for linking to the original designs, I obviously had forgotten about those @melchoyce.  Any design direction for front-end rendering or are we just using the core gallery rendering there?

melchoyce:
> Any design direction for front-end rendering or **are we just using the core gallery rendering there**?

Yup, since themes can style that.

wp-includes/widgets/class-wp-widget-media-gallery.php
timmyc: Strike that comment, I see the check further down now :)

timmyc:
@westonruter any ideas around the last error from CI?

`Invalid symmetric difference expression master`

timmyc:
I did some of the updates from my prior review, and added in qunit and phpunit test suites.  Both test suites could use some more coverage.

One thing I'm hoping we can attempt is to not have to serialize the entire attachment object for each image in the `attachments` data on the widget instance.  As-is there are things like nonces in the data, and it is really just a large set of data to be persisting... not to mention it will get out of date.

I know the gallery block in gutenberg was able to pare down this data quite a bit, so maybe we can take some pointers from the prior art there.  I attempted to just `_.pick` the `id,url,sizes` but that broke the selection state in the modal when editing an existing gallery.  I will try to dig in to this a bit more, but wanted to log here in case anyone else wanted to give it a try.

m1tk00:
@timmyc so what your suggesting is not storing the `attachment` data into the widget at all?

joemcgill:
While not completely finished yet, you can see in [this PR](https://github.com/WordPress/gutenberg/pull/2488) and example of how to use `wp.media.query` to fetch attachment models before setting up the media frame, which is the better way to avoid storing any data from the attachments besides the post IDs.

m1tk00:
@joemcgill I saw how the gallery shortcode javascript works and they seem to use the wp.media.query as well. Ive copied and modified the functionality to get the attachments details. The question that follows is how do we send the attachment data in the render functionality via php or calling the same functionality from the gallery class?

joemcgill:
@m1tk00 I'm not sure that I understand your question, so sorry if I don't address it correctly. Once attachments are selected from the media frame, I think the widget field should make use of the standard gallery shortcode, which will use the `gallery_shortcode()` function from WordPress to parse the shortcode attributes and render the correct HTML.

m1tk00:
@joemcgill I was referring to [this screen](https://cloud.githubusercontent.com/assets/2846578/25029086/e8b7efa2-2087-11e7-8ea6-43679a3a2fb8.png). At the moment this images come from the widget attributes attribute. The frontend goes with the `gallery_shortcode()` function

joemcgill:
@m1tk00 Got it. I'm not that familiar with the work that has been done for this widget thus far. Perhaps that's a better question for @timmyc or @westonruter.

timmyc:
Okay wanted to drop some notes here on some items that I think still need some attention on this branch:

- [ ] Currently CI is broken, bad on me for that
- [ ] We still need to implement the [enhanced widget preview design](https://github.com/xwp/wp-core-media-widgets/issues/62#issuecomment-294054977)
- [ ] Right now the widget instance stores all media attributes serialized - a snapshot I should say - when an image is selected for the gallery. This is problematic because the data can get out of synch from the media library data, and it just really bloats the image instance data itself. I think a good guideline here would be to aim for what the gutenberg block `save`s for an instance, the media `ID`s and urls seem sufficient for our needs:

```
<!-- wp:core/gallery -->
<div class="wp-block-gallery alignnone columns-3 is-cropped">
    <figure class="blocks-gallery-image"><img src="http://haiku2.com/wp-content/uploads/2017/08/Add_Plugins_‹_haiku2_—_WordPress.png" alt="" data-id="493" /></figure>
    <figure class="blocks-gallery-image"><img src="http://haiku2.com/wp-content/uploads/2016/10/A_topnotch_WordPress_com_site-2-1.png" alt="" data-id="305" /></figure>
    <figure class="blocks-gallery-image"><img src="http://haiku2.com/wp-content/uploads/2016/10/prince_of_pancakes_by_ucarts-d7tav5q.jpg" alt="" data-id="301" /></figure>
</div>
<!-- /wp:core/gallery -->
```

westonruter:
@timmyc shouldn't the attachment IDs be all that is required to store here? The `gallery_shortcode` function only recognizes an array of attachment IDs anyway (the `ids` shortcode attribute). There isn't a way currently to create a gallery in core that is made up of external images.

westonruter:
Also, the build error is a glitch with Travis and/or wp-dev-lib: “fatal: Invalid symmetric difference expression master”. So it's not your bad. Pushing another commit to the branch normally clears that up. It's a difficult issue to debug. Tried to fix in https://github.com/xwp/wp-dev-lib/pull/222 but it hasn't been successful.

timmyc:
>  shouldn't the attachment IDs be all that is required to store here?

I suppose that is correct, as long as we can use existing media library tooling to grab the thumbnails to build the widget preview we should be okay with just IDs

wp-includes/widgets/class-wp-widget-media-gallery.php
```diff
@@ -0,0 +1,226 @@
...
+							<dt class="gallery-icon">
+							<# if ( attachment.sizes.thumbnail ) { #>
+								<img src="{{ attachment.sizes.thumbnail.url }}" width="{{ attachment.sizes.thumbnail.width }}" height="{{ attachment.sizes.thumbnail.height }}" alt="" />
+							<# } else { #>
+								<img src="{{ attachment.url }}" alt="" />
+							<# } #>
+							</dt>
+							<# if ( attachment.caption ) { #>
+								<dd class="wp-caption-text gallery-caption">
+									{{{ data.verifyHTML( attachment.caption ) }}}
```
westonruter: This line is throwing an error for me: “data.verifyHTML is not a function”. It is `undefined`. What is this function supposed to do?

westonruter (dismissed):

m1tk00:
@timmyc you forgot to mention the issue when updating the gallery settings. Im working on that now along with the design for the previewer.

joemcgill:
@m1tk00 I've noticed issues updating settings as well while working on #187. Do we have a list of things that need to be fixed to make sure we address all of them? Also, I started taking a crack at the new preview design as well and opened a WIP PR #188.

wp-includes/widgets/class-wp-widget-media-gallery.php
joemcgill: This may become moot based on the updated widget preview design. See #188.

wp-includes/widgets/class-wp-widget-media-gallery.php
westonruter: I think this would be handled automatically by the fact that if you pass a deleted ID into `gallery_shortcode()` it will just try to include that ID in the query but if it doesn't match then it simply won't be included in the results. This works since we know an attachment could be deleted but it doesn't cause a containing gallery on a post to fail.

tests/qunit/test-media-image-widget.js
```diff
@@ -31,7 +31,9 @@
 		equal( templateProps.currentFilename, 'wp-header-logo.png', 'mapModelToPreviewTemplateProps should set currentFilename based off of url' );
 		equal( templateProps.url, testImageUrl, 'mapModelToPreviewTemplateProps should return the proper url' );
 		equal( templateProps.alt, 'some alt text', 'mapModelToPreviewTemplateProps should return the proper alt text' );
-		equal( templateProps.link_type, undefined, 'mapModelToPreviewTemplateProps should ignore attributes that are not needed in the preview' );
+
+		// TODO Fix this upstream.
+		//equal( templateProps.link_type, 'none', 'mapModelToPreviewTemplateProps should ignore attributes that are not needed in the preview' );
```
westonruter, line 36: What is the patch needed upstream?

westonruter:
📣 Anything key outstanding for this PR to hold it back from merge and initial core commit?

joemcgill:
@westonruter last I checked, I don't think gallery settings from the media frame, like columns, etc,  are being saved to the widget. I was going to sweep through and test each setting to see what was left.

wp-admin/js/widgets/media-gallery-widget.js
```diff
@@ -0,0 +1,297 @@
...
+		 *
+		 * @returns {void}
+		 */
+		renderPreview: function renderPreview() {
+			var control = this, previewContainer, previewTemplate, attachments;
+
+			previewContainer = control.$el.find( '.media-widget-preview' );
+			previewTemplate = wp.template( 'wp-media-widget-gallery-preview' );
+
+			if ( control.model.get( 'ids' ).length && ! control.model.get( 'attachemnts' ) ) {
```
westonruter: `! control.model.get( 'attachemnts' )` this looks like it can be eliminated. At least, it is misspelled.

wp-admin/js/widgets/media-gallery-widget.js
```diff
@@ -0,0 +1,297 @@
...
+		 * @returns {void}
+		 */
+		removeAttachmentId: function removeAttachmentId( id ) {
+			var attachments, newAttachments;
+			attachments = JSON.parse( this.get( 'attachments' ) );
+			newAttachments = _.filter( attachments, function( attachment ) {
+				return attachment.id !== id;
+			} );
+			this.set( {
+				'attachments': JSON.stringify( newAttachments ),
```
westonruter: I think references to 'attachments' still needs to be eliminated.

wp-admin/js/widgets/media-gallery-widget.js
```diff
@@ -0,0 +1,297 @@
...
+				'ids': _.map( newAttachments, 'id' )
```
westonruter: I think this actually should be `_.pluck( newAttachments, 'id' )` not `map`.

westonruter:
I'm working on some improvements to the rendering and syncing of the gallery settings.

wp-admin/js/widgets/media-gallery-widget.js
```diff
@@ -0,0 +1,302 @@
...
+				selection = new wp.media.model.Selection( control.model.get( 'attachments' ), {
+					multiple: true
+				});
+			} else {
+				selection = null;
+			}
+
+			mediaFrameProps = control.mapModelToMediaFrameProps( control.model.toJSON() );
+			if ( mediaFrameProps.size ) {
+				control.displaySettings.set( 'size', mediaFrameProps.size );
```
westonruter: This doesn't seem to be used?

wp-admin/js/widgets/media-gallery-widget.js
```diff
@@ -0,0 +1,302 @@
...
+			var control = this, selection, mediaFrame, defaultSync, mediaFrameProps;
+			if ( control.isSelected() && 0 !== control.model.get( 'selection' ) ) {
+				selection = new wp.media.model.Selection( [ control.selectedAttachment ] );
+			} else {
+				selection = null;
+			}
+
+			mediaFrameProps = control.mapModelToMediaFrameProps( control.model.toJSON() );
+			if ( mediaFrameProps.size ) {
+				control.displaySettings.set( 'size', mediaFrameProps.size );
```
westonruter: This doesn't seem to be used?

wp-admin/js/widgets/media-gallery-widget.js
westonruter: It was needed. I confirmed.

westonruter:
@joemcgill there's one key thing thing left before I think we can call this ready to merge: customizations.

When you have selected images for a gallery, you then see something like this:

image

If you try supplying a caption then:

image

And then it update, there is no changes made because there is no `attachments` in the model there is nowhere for the customizations to go. Normally in the post editor when you edit a caption it then gets written back to the underlying attachment. However, in the case of media widgets this is blocked entirely with this logic:

```js
// Disable syncing of attachment changes back to server (except for deletions). See <https://core.trac.wordpress.org/ticket/40403>.
defaultSync = wp.media.model.Attachment.prototype.sync;
wp.media.model.Attachment.prototype.sync = function( method ) {
	if ( 'delete' === method ) {
		return defaultSync.apply( this, arguments );
	} else {
		return $.Deferred().rejectWith( this ).promise();
	}
};
mediaFrame.on( 'close', function onClose() {
	wp.media.model.Attachment.prototype.sync = defaultSync;
});
```

It seems that for the Gallery widget we must remove this logic or else a user will not be able to modify the image attributes.

joemcgill:
@westonruter I was looking at that syncing logic yesterday and agree with you that in this context, captions should be able to be synced back to the underlying attachment. Your updates to the syncing of props between the widget control and the media frame seem to fix the issues I had noticed but the gallery attributes weren't being used to render the gallery shortcode on the front end, so I pushed another quick commit to resolve that issue.

Will do some more testing, but other than the caption issue, things are looking good from my end. Just a note that captions are a bit tricky because some themes disregard captions when they render galleries.

westonruter (approved):

westonruter:
@joemcgill excellent, thanks. Once you merge this PR then we'll release a new version of the feature plugin to get out to testers for next week prior to core merge.

joemcgill:
@westonruter Got a chance to this again after e5ee972, and captions updates are now saving back to the model correctly. Going to go ahead and merge so we can get this into more hands. Nice work 🎉

westonruter:
Great work everyone!

westonruter:
The plugin auto-deployed to WordPress.org so it is now available for all to test: https://wordpress.org/plugins/wp-core-media-widgets/

Issue #126, opened 2017-05-04
Video sources are not reset when selecting external video embed


westonruter:
1. Add video from media library.
2. Select alternative sources
3. Select an external video via YouTube or Vimeo
4. Open Edit Video frame.
5. Unexpectedly see the original alternative sources from the previous video still selected.

The video sources should get cleared when selecting an embed. We should extend `getEmbedResetProps` in `VideoWidgetControl` to include resets for all of the video extension props.

westonruter:
Related to #99

westonruter:
Oh, another complication with `url` vs video extension source props… in the post editor, if I have a video shortcode that has both a `url` and an `mp4` pointing to the same URL, and then I go in and replace the video, the `url` prop persists and there is a conflict in the state.

Maybe we should only allow using the `url` prop for external video embeds like from YouTube and Vimeo, and _always_ use the video extension props for videos uploaded from the media library. When selecting an embed, all of the video props should get reset. And when selecting a video from the media library, the `url` prop should get reset.

westonruter:
Fixed in #127.

PR #127, opened 2017-05-04
Add video/audio extension awareness

obenland:
Adds a widget-specific way to determine whether they have content or not and populates all extensions based on their values.

### DIFF
--- a/tests/qunit/test-media-image-widget.js
+++ b/tests/qunit/test-media-image-widget.js
@@ -62,9 +62,9 @@
 		// Test mapMediaToModelProps
 		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'file', url: testImageUrl } );
 		equal( mappedProps.link_url, testImageUrl, 'mapMediaToModelProps should set file link_url according to mediaFrameProps.link' );
-		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'post', postUrl: 'https://wordpress.org/image-2/' } );
+		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'post', postUrl: 'https://wordpress.org/image-2/', url: testImageUrl } );
 		equal( mappedProps.link_url, 'https://wordpress.org/image-2/', 'mapMediaToModelProps should set file link_url according to mediaFrameProps.link' );
-		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'custom', linkUrl: 'https://wordpress.org' } );
+		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'custom', linkUrl: 'https://wordpress.org', url: testImageUrl } );
 		equal( mappedProps.link_url, 'https://wordpress.org', 'mapMediaToModelProps should set custom link_url according to mediaFrameProps.linkUrl' );
 
 		// Test mapModelToMediaFrameProps().
--- a/tests/qunit/test-media-image-widget.js
+++ b/tests/qunit/test-media-image-widget.js
@@ -66,6 +66,7 @@
 		equal( mappedProps.link_url, 'https://wordpress.org/image-2/', 'mapMediaToModelProps should set file link_url according to mediaFrameProps.link' );
 		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'custom', linkUrl: 'https://wordpress.org', url: testImageUrl } );
 		equal( mappedProps.link_url, 'https://wordpress.org', 'mapMediaToModelProps should set custom link_url according to mediaFrameProps.linkUrl' );
+		equal( mappedProps.png, undefined, 'mapMediaToModelProps should not set the extension on an image' );
 
 		// Test mapModelToMediaFrameProps().
 		imageWidgetControlInstance.model.set({ error: false, url: testImageUrl, 'link_type': 'custom', 'link_url': 'https://wordpress.org', 'size': 'custom', 'width': 100, 'height': 150, 'title': 'widget title', 'image_title': 'title of image' });
--- a/tests/qunit/test-media-image-widget.js
+++ b/tests/qunit/test-media-image-widget.js
@@ -62,11 +62,10 @@
 		// Test mapMediaToModelProps
 		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'file', url: testImageUrl } );
 		equal( mappedProps.link_url, testImageUrl, 'mapMediaToModelProps should set file link_url according to mediaFrameProps.link' );
-		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'post', postUrl: 'https://wordpress.org/image-2/', url: testImageUrl } );
+		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'post', postUrl: 'https://wordpress.org/image-2/' } );
 		equal( mappedProps.link_url, 'https://wordpress.org/image-2/', 'mapMediaToModelProps should set file link_url according to mediaFrameProps.link' );
-		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'custom', linkUrl: 'https://wordpress.org', url: testImageUrl } );
+		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'custom', linkUrl: 'https://wordpress.org' } );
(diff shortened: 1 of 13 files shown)


obenland:
One thing missing here is to populate those extensions from the PHP side on first load

wp-admin/js/widgets/media-widgets.js
```diff
@@ -713,6 +713,20 @@ wp.mediaWidgets = ( function( $ ) {
...
+			// Reset all video/audio sources.
+			wp.media.view.settings.embedExts.map( function( ext ) {
+
+				// Only reset when the main source changes.
+				if ( ! ( ext in modelProps ) || isEmbed ) {
+					modelProps[ ext ] = '';
+				}
+			} );
+
+			extension = mediaFrameProps.url.split( '.' ).pop();
```
timmyc: Do we need to be concerned about case here, i.e. if a file has an extension of `.MP4` instead of `.mp4` maybe call `toLowerCase()` on it just to be safe?

wp-admin/js/widgets/media-widgets.js
```diff
@@ -713,6 +713,20 @@ wp.mediaWidgets = ( function( $ ) {
...
+			modelProps[ extension ] = mediaFrameProps.url;
```
timmyc: Should we check to see if `wp.media.view.settings.embedExts[ extension ]` exists first?  As-is won't this set a prop of `jpg` on image widgets?

tests/qunit/test-media-image-widget.js
```diff
@@ -62,9 +62,9 @@
 		// Test mapMediaToModelProps
 		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'file', url: testImageUrl } );
 		equal( mappedProps.link_url, testImageUrl, 'mapMediaToModelProps should set file link_url according to mediaFrameProps.link' );
-		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'post', postUrl: 'https://wordpress.org/image-2/' } );
+		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'post', postUrl: 'https://wordpress.org/image-2/', url: testImageUrl  } );
 		equal( mappedProps.link_url, 'https://wordpress.org/image-2/', 'mapMediaToModelProps should set file link_url according to mediaFrameProps.link' );
-		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'custom', linkUrl: 'https://wordpress.org' } );
+		mappedProps = imageWidgetControlInstance.mapMediaToModelProps( { link: 'custom', linkUrl: 'https://wordpress.org', url: testImageUrl  } );
```
timmyc: Should we add some more tests here to cover the extensions being set/reset?

timmyc (commented):
I used the steps outlined in #126 and verified that the alternate sources are no longer showed when switching from a local file -> YouTube video.  Few other minor comments.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -420,4 +402,38 @@ public function render_control_template_scripts() {
...
+	 * Creates and returns a link for an attachment.
+	 *
+	 * @since 4.8.0
+	 * @access protected
+	 *
+	 * @param WP_Post $attachment Attachment object.
+	 * @param string  $type       link type.
+	 * @return string
+	 */
+	protected function create_link_for( $attachment, $type = '' ) {
```
westonruter: I think this method can be just removed entirely, right? It's not even used.

wp-admin/js/widgets/media-widgets.js
westonruter: Shouldn't this only be done if `isEmbed`? Also, if `! isEmbed` then I believe that `modelProps.url = ''` needs to be done, so that we reserve `url` for external embeds and then the extension props for media uploaded to the media library?

obenland: It would, but the schema takes care of that for us and only adopts whitelisted values

tests/qunit/test-media-image-widget.js
obenland: Absolutely we should

wp-includes/widgets/class-wp-widget-media.php
obenland: Ha! Even better. I didn't check

wp-admin/js/widgets/media-widgets.js
obenland: If we do it for uploaded files as well, we'll automatically populate the additional sources part in the edit modal.

I'm not sure I understand why you want to move away from uploaded files using `url`. Should that be set no matter what? When would we encounter an attachment without it?

westonruter: @obenland So you're fine with there being a `url` and `mp4` props that have the same value?

obenland: Unless it introduces unexpected behavior or errors, I don't see what the harm is. And we can continue to rely on `url` in all cases

westonruter: OK, sure. And I guess there is a slight advantage there because it allows the user to indicate the “primary” source in a sense.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -713,6 +713,20 @@ wp.mediaWidgets = ( function( $ ) {
...
+			wp.media.view.settings.embedExts.map( function( ext ) {
+
+				// Only reset when the main source changes.
+				if ( ! ( ext in modelProps ) || isEmbed ) {
+					modelProps[ ext ] = '';
+				}
+			} );
+
+			extension = mediaFrameProps.url.split( '.' ).pop().toLowerCase();
+			modelProps[ extension ] = mediaFrameProps.url;
```
westonruter: Doesn't this need to be done only if `! isEmbed`? In other words, if I select a YouTube video, then `extension` will not be valid.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -420,4 +402,17 @@ public function render_control_template_scripts() {
...
 	}
+
+	/**
+	 * Whether the widget has content to show.
+	 *
+	 * @since 4.8.0
+	 * @access protected
+	 *
+	 * @param array $instance Widget instance props.
+	 * @return bool
```
westonruter: Add return description.

wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -420,4 +402,17 @@ public function render_control_template_scripts() {
...
+	 */
+	protected function has_content( $instance ) {
+		return ( $instance['attachment_id'] && 'attachment' !== get_post_type( $instance['attachment_id'] ) ) || $instance['url'];
```
westonruter: Not sure I understand the logic in this. Shouldn't it include `'attachment' === get_post_type( $instance['attachment_id'] )` (`===` not `!==`)?

obenland: Yes, good catch! Missed that one in my conversion

wp-admin/js/widgets/media-widgets.js
obenland: Oy, I've only tested it with file urls so far

obenland: If `extension` is not valid it will be purged by the instance schema. 

Do you want us to be more verbose here and test against an extension whitelist?

westonruter: I'm more interested in the case where `mediaFrameProps.url` is equal to `https://www.youtube.com/watch?v=HTgYHHKs0Zw`, as in that case the “extension” will be `com/watch?v=HTgYHHKs0Zw`. So yeah, I guess we should skip this entirely if it doesn't have a file extension. Also, I think we should perhaps explicitly strip off any query or hash fragments from the URL so that if someone supplies a file URL with any additional params, the extension will get picked up properly.

In other words, something like:

```js
extension = mediaFrameProps.url.replace( /#.*$/, '' ).replace( /\?.*$/, '' ).split( '.' ).pop().toLowerCase()
```

And then yes, I think it would probably be good form to check if `extension` exists among `control.model.schema`.

timmyc:
Query string removal, extension check, and a few more tests added in fb597c8

westonruter:
It seemed to make more sense to reset the media source props in `getModelPropsFromMediaFrame` when we know that a _new_ media item was selected. Otherwise, `mapMediaToModelProps` shouldn't be resetting props when returning from the edit frame. I did this in fe9d590 though sadly this means some of the new tests can't still apply at `getModelPropsFromMediaFrame` is not testable since it relies on the media manager JS.

westonruter (approved):

Issue #130, opened 2017-05-05
video looping


soderlind:
Adding a [youtube video](https://www.youtube.com/watch?v=GPTY6l_PX5k) without doing any changes. Playing it, it loops at the end of the video. 
If I  edit the video and check "loop", it doesn't loop anymore, not even if I uncheck/check and save multiple times.

timmyc:
totally not related to the bug report, but that cover/video is amazing 😮 !!

timmyc:
Was able to reproduce with multiple videos - the same experience is not encountered when embeding via the editor either.

timmyc:
So debugging a bit more, it seems to only be affecting YouTube videos.  Embedding a Vimeo video ( I used https://vimeo.com/190372437 ) and upon playing the video it did not loop.

soderlind:
`<offtopic>` 
AURORA has also done [a cover of "Half the World Away"](https://www.youtube.com/watch?v=SYKU7DfEJT4) by  Oasis .. and thanks for pointing me to Frightened Rabbit :)
`</offtopic>`

timmyc:
Love the 🎵  angle on this issue!

Anyhow, wanted to update the original issue here that the YouTube looping issue appears to be a bug in the underlying `mediaelement` library version in use in core - details and more discussion on the PR https://github.com/xwp/wp-core-media-widgets/pull/136#issuecomment-299519018

westonruter:
Re-opening because YouTube is not fixed yet, which apparently depends on https://core.trac.wordpress.org/ticket/39686

rafa8626:
Is there a sandbox where I can see this issue and assist you to fix this please? Is this for 4.x version?

westonruter:
@rafa8626 just install this plugin from GitHub.

rafa8626:
Once installed is that all I need to do? Excuse my lack of knowledge in Wordpress

rafa8626:
I tried to add a YouTube video through the Add Media button, but when I paste the URL it has all YouTube styles, not MediaElement. Should I be doing this that way? If not, please indicate how so I can assist you and continue to move towards 4.x version. BTW, is this issue on 4.x version, or on 2.x version?

rafa8626:
I have submitted a patch on Wordpress ticket. Can you integrate it on your branch and confirm it's fixed, please?

PR #134, opened 2017-05-05
Prevent Title Loss When Editing Video

timmyc:
While looking at #130 I noted another flow that was causing the title of the video widget to be lost:

__Steps to Recreate__
- Add a video widget, set a title
- Edit the video widget to adjust `loop` value for example
- Note the widget title has been reset to the default empty string

Checkout this branch and note the steps above do not cause the title of the video widget to be reset to the default value.


obenland:
Good catch!

obenland (approved):

PR #135, opened 2017-05-05
Whitelist should_preview_update for video/audio

obenland:
Prevents the preview from updating when certain props change.

__Steps to Recreate__
- Add a video widget.
- Start entering a title for the widget.
- Note the preview updating with every key stroke.

Checkout this branch and note the steps above do not cause the preview to be refreshed.

### DIFF
--- a/wp-includes/widgets/class-wp-widget-media-audio.php
+++ b/wp-includes/widgets/class-wp-widget-media-audio.php
@@ -151,7 +151,7 @@ class WP_Widget_Media_Audio extends WP_Widget_Media {
 
 		$exported_schema = array();
 		foreach ( $this->get_instance_schema() as $field => $field_schema ) {
-			$exported_schema[ $field ] = wp_array_slice_assoc( $field_schema, array( 'type', 'default', 'enum', 'minimum', 'format', 'media_prop' ) );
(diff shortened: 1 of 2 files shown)


wp-includes/widgets/class-wp-widget-media-audio.php
```diff
@@ -151,7 +151,7 @@ public function enqueue_admin_scripts() {
 
 		$exported_schema = array();
 		foreach ( $this->get_instance_schema() as $field => $field_schema ) {
-			$exported_schema[ $field ] = wp_array_slice_assoc( $field_schema, array( 'type', 'default', 'enum', 'minimum', 'format', 'media_prop' ) );
+			$exported_schema[ $field ] = wp_array_slice_assoc( $field_schema, array( 'type', 'default', 'enum', 'minimum', 'format', 'media_prop', 'should_preview_update' ) );
```
timmyc, line 154: Thanks for catching this... do you think we should have a base array of exported fields that should be included on each class?  I suppose not that critical but just a thought.

timmyc (approved):
Thanks for fixing this - working as expected now.

wp-includes/widgets/class-wp-widget-media-audio.php
obenland: Yeah I was thinking about that too. I mean since `should_preview_update` is part of the base schema too it would make sense to have that included by default. But it does add more complexity. Not sure

PR #136, opened 2017-05-05
Fix Video Loop Issues

### LINKED ISSUE #130: video looping

timmyc:
This is still in progress as I still have yet to find a solution for the YouTube looping.  This branch seeks to fix #130 - though.  While testing out the original report, I noticed that the loop issue was only happening on YouTube embeds, and not on Vimeo embeds.  But even when `loop` was true for Vimeo embeds - they were not looping in the rendered widget.

A fix for the vimeo issue can be seen in dddaa88 ( which apparently has made CI unhappy ).  By adding the `loop` query arg on the Vimeo url, the underlying iframe from Vimeo loops when expected to.  YouTube also supports this query arg ( https://developers.google.com/youtube/player_parameters ), but it appears that mediaelement is mangling with the YouTube embed url... because even when the logic in this branch adds the `loop` query arg appropriately, the iframe src that is subsequently rendered in the widget no longer has the argument:

If I manually add the `loop` query arg back into the iframe URL seen above, the YouTube video no longer loops.

### DIFF
--- a/wp-includes/widgets/class-wp-widget-media-video.php
+++ b/wp-includes/widgets/class-wp-widget-media-video.php
@@ -115,7 +115,10 @@ class WP_Widget_Media_Video extends WP_Widget_Media {
 		if ( $attachment ) {
 			$src = wp_get_attachment_url( $attachment->ID );
 		} else {
-			$src = $instance['url'];
+
+			// Manually add the loop query argument.
+			$loop = $instance['loop'] ? '1' : '0';
+			$src = add_query_arg( 'loop', $loop, $instance['url'] );
 		}
 
 		if ( empty( $src ) ) {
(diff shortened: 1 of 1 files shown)


timmyc:
Related: https://github.com/mediaelement/mediaelement/issues/1851

and fixed in https://github.com/instructure/mediaelement/pull/17

And released in version 2.23.2 per https://github.com/instructure/mediaelement/blob/master/changelog.md

Version of mejs in trunk is 2.22.0

obenland:
Maybe we should update the library in 4.8?

timmyc:
I think it was talked about in the bug scrub today: https://core.trac.wordpress.org/ticket/39686

The issue is to upgrade to a new major version release, so not sure if we can push to do a minor bump to 2.x or not... fwiw any use of `wp_video_shortcode` with a YouTube video is also affected by this.  I need to circle back and see if we can mimic the editor embed flow instead since it isn't looping.

obenland:
With mediaelement updated, would we still need this proposed change?

westonruter (approved):

PR #138, opened 2017-05-05
Prevent rebuilding editor upon save by moving fields outside of 'managed' element

### LINKED ISSUE #132: Text Widget: why destroy tinymce instance?
Wonder what's the reasoning for needing to destroy tinymce after widget updates. Also, it may be worth to move this block of code into a separate "prepareEditor" function or similar.

https://github.com/xwp/wp-core-media-widgets/blob/master/wp-admin/js/widgets/text-widgets.js#L60-L76

westonruter:
Fixes #132.

### DIFF
--- a/wp-admin/js/widgets/text-widgets.js
+++ b/wp-admin/js/widgets/text-widgets.js
@@ -50,7 +50,7 @@ wp.textWidgets = ( function( $ ) {
 			id = textarea.attr( 'id' );
 
 			/**
-			 * Build (or re-build) an the visual editor.
+			 * Build (or re-build) the visual editor.
 			 *
 			 * @returns {void}
 			 */
--- a/wp-admin/js/widgets/text-widgets.js
+++ b/wp-admin/js/widgets/text-widgets.js
@@ -32,21 +32,83 @@ wp.textWidgets = ( function( $ ) {
 		initialize: function initialize( options ) {
 			var control = this;
 
+			if ( ! options.el ) {
+				throw new Error( 'Missing options.el' );
+			}
+
 			Backbone.View.prototype.initialize.call( control, options );
 
-			if ( ! control.el ) {
-				throw new Error( 'Missing options.el' );
+			/*
+			 * Create a container element for the widget control fields.
+			 * This is inserted into the DOM immediately before the the .widget-content
+			 * element because the contents of this element are essentially "managed"
+			 * by PHP, where each widget update cause the entire element to be emptied
+			 * and replaced with the rendered output of WP_Widget::form() which is
+			 * sent back in Ajax request made to save/update the widget instance.
+			 * To prevent a "flash of replaced DOM elements and re-initialized JS
+			 * components", the JS template is rendered outside of the normal form
+			 * container.
+			 */
+			control.fieldContainer = $( '<div class="text-widget-fields"></div>' );
+			control.fieldContainer.html( wp.template( 'widget-text-control-fields' ) );
+			control.widgetContentContainer = control.$el.find( '.widget-content:first' );
+			control.widgetContentContainer.before( control.fieldContainer );
+
+			control.fields = {
+				title: control.fieldContainer.find( '.title' ),
+				text: control.fieldContainer.find( '.text' )
+			};
+
+			// Sync input fields to hidden sync fields which actually get sent to the server.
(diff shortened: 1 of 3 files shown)


wp-includes/widgets/class-wp-widget-visual-text.php
```diff
@@ -149,18 +154,31 @@ public function form( $instance ) {
 				'text' => '',
 			)
 		);
+		?>
+		<input id="<?php echo $this->get_field_id( 'title' ); ?>" name="<?php echo $this->get_field_name( 'title' ); ?>" class="title" type="hidden" value="<?php echo esc_attr( $instance['title'] ); ?>">
+		<input id="<?php echo $this->get_field_id( 'text' ); ?>" name="<?php echo $this->get_field_name( 'text' ); ?>" class="text" type="hidden" value="<?php echo esc_attr( $instance['text'] ); ?>">
+		<?php
+	}
 
-		$title = sanitize_text_field( $instance['title'] );
```
westonruter, line 153: This was redundant so I removed it. The title is sanitized in the `update` method. All we need here is outputting the value with the appropriate escaping.

wp-includes/widgets/class-wp-widget-visual-text.php
```diff
@@ -149,18 +154,31 @@ public function form( $instance ) {
 				'text' => '',
 			)
 		);
+		?>
+		<input id="<?php echo $this->get_field_id( 'title' ); ?>" name="<?php echo $this->get_field_name( 'title' ); ?>" class="title" type="hidden" value="<?php echo esc_attr( $instance['title'] ); ?>">
+		<input id="<?php echo $this->get_field_id( 'text' ); ?>" name="<?php echo $this->get_field_name( 'text' ); ?>" class="text" type="hidden" value="<?php echo esc_attr( $instance['text'] ); ?>">
```
westonruter, line 159: There is a chance that existing plugins that try to extend the Text widget could trip over this here. If they try to obtain a textarea in here, they will fail. But then again, the textarea's behavior is radically changed with the introduction of TinyMCE, so perhaps it's a good thing that it is moved out of the way to isolate it.

wp-admin/js/widgets/text-widgets.js
```diff
@@ -32,25 +32,87 @@ wp.textWidgets = ( function( $ ) {
...
+				control.fields.title.val( syncInput.val() );
+			}
+
+			syncInput = control.widgetContentContainer.find( 'input[type=hidden].text' );
+			if ( control.fields.text.is( ':visible' ) ) {
+				if ( ! control.fields.text.is( document.activeElement ) ) {
+					control.fields.text.val( syncInput.val() );
+				}
+			} else if ( control.editor && ! control.editorFocused && syncInput.val() !== control.fields.text.val() ) {
+				control.editor.setContent( wp.editor.autop( syncInput.val() ) );
```
timmyc, line 100: 🤔  could a user make a change while the request is in-flight, then un-focus mce, and potentially lose the changes here?

westonruter: Yes, that seems like it could happen. Humm. I suppose we'd need to keep track of `change` events in addition to whether the editor is currently focused to determine whether we should override the content?

timmyc: I suppose it comes down to what we should consider the source of truth in the flow - it seems like we would always want the content in the widget to be preserved in this scenario, because wouldn't the final 'save' action perform any further sanitization needed?

timmyc: I can't get it to fail locally, but then again my response time on the admin-ajax call is pretty dang quick 😆

westonruter: Try adding `sleep( 3 )` to `admin-ajax.php`

westonruter: @timmyc In the customizer, there is already logic in place in core to prevent updates from overriding content that the user is entering: https://github.com/WordPress/wordpress-develop/blob/656718a645632771207b1ecb54a0c49e4b06033a/src/wp-admin/js/customize-widgets.js#L1221-L1227

On the widgets admin screen, on the other hand, it is already “the way it works” for fields to be blown away after hitting save. So I don't think we're introducing anything new.

timmyc (approved):
On that note :shipit:

PR #139, opened 2017-05-05
Restore editor to text mode after move

### LINKED ISSUE #137: Text (HTML) tab in TinyMCE editor should get focus after rebuild
If a user selects the Text tab in the editor, and then moves the widget to a new position in the sidebar, then when they expand the widget again they will see the Visual tab has been selected instead of the Text tab. We should remember the tab that was selected and restore it.

Complication: `wp.editor.initialize()` does not currently allow for the editor to be initialized with the Text tab selected.

timmyc:
This branch fixes #137.  The proposed solution here is to track wether or not the current mce instance `isHidden()` prior to being removed, and if it was `hidden`/ in text mode, utilize `switchEditors.go` to toggle the editor ( thanks for the tip @azaozz ! )

__To Test__
- Add a text widget
- Add some content, toggle to the Text tab
- Move the widget 
- Verify the freshly rendered MCE instance has the Text tab toggled on

### DIFF
--- a/wp-admin/js/widgets/text-widgets.js
+++ b/wp-admin/js/widgets/text-widgets.js
@@ -1,4 +1,4 @@
-/* global tinymce, QTags */
+/* global tinymce, QTags, switchEditors */
 /* eslint consistent-this: [ "error", "control" ] */
 wp.textWidgets = ( function( $ ) {
 	'use strict';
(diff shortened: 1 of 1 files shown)


westonruter (approved):
Works!

PR #146, opened 2017-05-08
Remove attachment title as img[title] fallback for image_title

### LINKED ISSUE #143: Title attribute in Image widget
I think title attribute should not be output by default when using Image widget. 

It should work the same way as in post.

1. Add image in a post and there is no title attribute from the Title input field.
1. Edit the image and under Advanced Options >> Image Title Attribute you can add title attribute if needed (not sure should there ever be a reason to add title attribute anyways).

@afercia do you have opinion on this one?

westonruter:
Fixes #143.


timmyc (approved):
LGTM

PR #155, opened 2017-05-08
Fix margins for media and buttons in control

### LINKED ISSUE #148: Placeholder spacing issues
Reported by @celloexpressions:

> Incorrect spacing between the placeholder and the buttons (2em total, instead of 1em = 13px). Remove top margin from buttons and add bottom margin on preview, or remove bottom margin on placeholder to fix.

Confirmed this is the case. Looks like the HTML structure is different for the existing media elements in the Customizer, and the widgets. For example, here's the header image:

```html
<div class="attachment-media-view">
    <div class="placeholder">No video selected</div>
    ...
</div>
```

Here's the image widget:

```html
<div class="attachment-media-view">
   <p class="placeholder">No image selected</p>
</div>
```

Since `.placeholder` is applied to a `<p>` instead of a `<div>`, it inherits extra margins from the default `<p>` styling. We should change this to a `<div>` to match the existing structure.

That said, I'd like to see about tweaking these styles Customizer-wide in core, since I do think that the lack of margin makes the placeholder feel cramped.

westonruter:
Fixes #148.

See:

Without media selected

With media selected

### DIFF
--- a/wp-admin/js/widgets/media-image-widget.js
+++ b/wp-admin/js/widgets/media-image-widget.js
@@ -31,6 +31,10 @@
 		 */
 		renderPreview: function renderPreview() {
 			var control = this, previewContainer, previewTemplate;
+			if ( ! control.model.get( 'attachment_id' ) && ! control.model.get( 'url' ) ) {
+				return;
+			}
+
 			previewContainer = control.$el.find( '.media-widget-preview' );
 			previewTemplate = wp.template( 'wp-media-widget-image-preview' );
 			previewContainer.html( previewTemplate( _.extend( control.previewTemplateProps.toJSON() ) ) );
--- a/wp-includes/widgets/class-wp-widget-media-image.php
+++ b/wp-includes/widgets/class-wp-widget-media-image.php
@@ -317,10 +317,6 @@ class WP_Widget_Media_Image extends WP_Widget_Media {
 						echo sprintf( __( 'Current image: %s' ), '{{ data.currentFilename }}' );
 					?></p>
 				<# } #>
-			<# } else { #>
-				<div class="attachment-media-view">
-					<p class="placeholder"><?php echo esc_html( $this->l10n['no_media_selected'] ); ?></p>
-				</div>
 			<# } #>
 		</script>
 		<?php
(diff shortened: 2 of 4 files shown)


wp-admin/js/widgets/media-image-widget.js
```diff
@@ -31,6 +31,10 @@
 		 */
 		renderPreview: function renderPreview() {
 			var control = this, previewContainer, previewTemplate;
+			if ( ! control.model.get( 'attachment_id' ) && ! control.model.get( 'url' ) ) {
```
obenland, line 34: Should we use `isSelected()` here?

wp-includes/widgets/class-wp-widget-media-image.php
```diff
@@ -317,10 +317,6 @@ public function render_control_template_scripts() {
 						echo sprintf( __( 'Current image: %s' ), '{{ data.currentFilename }}' );
 					?></p>
 				<# } #>
-			<# } else { #>
-				<div class="attachment-media-view">
-					<p class="placeholder"><?php echo esc_html( $this->l10n['no_media_selected'] ); ?></p>
```
obenland, line 322: There is never an instance when we'd update the preview without an image, correct?

westonruter: That's the idea yes. You'll see that this change brings the image widget in line with the video and audio widgets, where they too lack this `else` statement.

wp-admin/js/widgets/media-image-widget.js
westonruter: Great question. I copied this statement from the `renderPreview` method in the video and audio widgets. Perhaps they should use `isSelected` as well…

wp-includes/widgets/class-wp-widget-media-image.php
westonruter: Actually, no. You can see that `isSelected` is looking at `control.model.get( 'error' )` and in this case here we need to pass the `error` into the template.

obenland: Good point

obenland (approved):

PR #156, opened 2017-05-08
Fix width of attachment filters container

### LINKED ISSUE #153: Narrow date inputs in the media library
screen shot 2017-05-08 at 2 56 53 pm

From @celloexpressions:

> Ridiculously-narrow inputs (select for dates and search input) in .media-toolbar in the media modal. Not sure if it's specific to the widgets' use of the modal but shouldn't be particularly hard to fix.

I've seen this before with the cut-off date dropdown, but I remember it being fixed in a previous release, so this might be a regression. 

Search input is inline with the rest of the media modals, though, so if we want to change that it should happen for all of core.

westonruter:
Fixes #153 

For some reason the container is not wide enough when there is only one select drop-down filter to cause the right `calc()` computations, apparently.

Results:

image

### DIFF
--- a/wp-admin/css/widgets/media-widgets.css
+++ b/wp-admin/css/widgets/media-widgets.css
@@ -47,6 +47,11 @@
 .media-widget-control .media-widget-preview .wp-video-shortcode {
 	background: #000;
 }
+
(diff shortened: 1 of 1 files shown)


timmyc (approved):
Looks good.

PR #157, opened 2017-05-09
Improve Insert by URL frame for various media types

### LINKED ISSUE #142: Provide help text for inserting a YouTube video
> Displays a video from the media library or from YouTube, Vimeo, or another provider.

In Video widget it took me a while that I can enter youtube video by selecting **Insert from URL**. I was just wondering could there be some kind of help text for that when you select video widget. 

For example:

> Displays a video from the media library. Or from YouTube, Vimeo, or another provider by selecting Insert from URL.

I also opened another video related issue in #141.

### LINKED ISSUE #149: Detect invalid media type URL
From @celloexpressions:

> No warning or error when using the wrong type of URL for a given widget. I accidentally pasted a YouTube URL into the audio widget and there were no warnings or errors, but it then shows the link and "download file". There needs to be feedback about why it's not working for this situation or, even better, the widget should adapt to the appropriate widget type based on the content used. For people who frequently use YouTube for music, the audio widget seems to suggest that a YouTube URL would work but then it gives a very odd experience.

We should try to figure out if we can detect when someone is pasting in a URL for the wrong kind of media type. I'll design an error state and then we can explore more.

### LINKED ISSUE #171: .mov Video doesn't play on frontend
Meta boilerplate, much ❤️ for this. These widgets are ones I've been longing for for a decade. I'm not kidding.

The video widget works great in the admin:

video works great in admin

But it doesn't actually render on the frontend —  or rather, it becomes a link, when I would have expected a video player:

video works great in admin

Here's a snap of the markup:

markup

I've no opinion on what type of video player it should use, whether a fancy bundled JS one, or just the stock HTML5 browser provided one that's used in the admin. I personally think the latter is fine.

westonruter:
Fixes #149.
Fixes #171.

When an invalid URL is detected either by file extension or oEmbed response, the Add to Widget button is disabled and the URL field gets a red outline.

Attempting to add an MP3 file to a video widget:

image

Attempting to add a YouTube video to an audio widget:

image

Attempting to use a non-image URL for the Image widget shows a broken image now:

image


wp-admin/js/widgets/media-widgets.js
```diff
@@ -121,8 +121,34 @@ wp.mediaWidgets = ( function( $ ) {
...
 
+						fetchSuccess = function( response ) {
+							embedLinkView.renderoEmbed({
+								data: {
+									body: response
+								}
+							});
+
+							$( '#embed-url-field' ).removeClass( 'invalid' );
+							embedLinkView.views.parent.views.parent.views.get( '.media-frame-toolbar' )[0].$el.find( '.media-button-select' ).prop( 'disabled', false );
```
timmyc: Looks like this code is duplicated a few times in the PR here, would a helper method to toggle the button make sense?  Also would be really nice if there was an easier way to access these buttons and disable them.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -121,8 +121,34 @@ wp.mediaWidgets = ( function( $ ) {
...
+						};
+
+						urlParser = document.createElement( 'a' );
+						urlParser.href = embedLinkView.model.get( 'url' );
+						matches = urlParser.pathname.toLowerCase().match( /\.(\w+)$/ );
```
timmyc, line 171: I'm a little concerned about relying upon files on the web to always have extensions to test for mime compatibility here.

wp-admin/js/widgets/media-widgets.js
```diff
@@ -134,13 +160,13 @@ wp.mediaWidgets = ( function( $ ) {
 						});
 
 						embedLinkView.dfd.done( function( response ) {
-							embedLinkView.renderoEmbed( {
-								data: {
-									body: response.html
-								}
-							});
+							if ( embedLinkView.controller.options.mimeType !== response.type ) {
+								embedLinkView.renderFail();
```
timmyc, line 198: This will now `renderFail` on any service that doesn't have embed discovery - for certain any now that are not supported by noembed... which maybe is okay - but something we should consider

timmyc (commented):
I'm hoping some others can take this branch for a spin - it is working most times for me, but I am able to get things in a state where I'm still able to insert the incorrect type of media... and other times where I hit a `renderFail()` state - typically from the return at `if ( $( '#embed-url-field' ).val() !== embedLinkView.model.get( 'url' ) ) {` when I have a valid URL.

Part of me thinks `renderFail` should be spitting out a warning message, that gives the user a bit more information about why the preview failed, and how they can fix it - but perhaps we should always allow users to insert whichever URL they want.

I also noted some concerns about relying upon file extensions to determine file types... that can be problematic too.  I suppose not matter what we do here, there are going to be edge cases where users will put URLs in to the widgets that cause them to break.  To me the decision here is whether or not to introduce more complexity to try to block a percentage of those edge cases.

wp-admin/js/widgets/media-widgets.js
westonruter: YOU DON'T THINK THIS IS ELEGANT!?

😉

wp-admin/js/widgets/media-widgets.js
westonruter: Yeah, it's a valid concern. Nevertheless, file extensions are relied on in the video shortcode for rendering: https://github.com/WordPress/wordpress-develop/blob/c449b34f6c35e179925110e076797a9c7df4a7f2/src/wp-includes/media.php#L2494-L2498

So I don't think we're doing anything that isn't aligned with what core is already doing.

wp-admin/js/widgets/media-widgets.js
westonruter: In any case, the video widget only supports YouTube and Vimeo. The audio widget doesn't support oEmbed at all: you can _only_ provide URLs to audio files. This would need to be revisited for adding additional media types.

westonruter:
Beware I just rebased this branch.

westonruter:
Now when providing an invalid video URL:

image

timmyc (approved):
This is testing out well for me, and think these are some great safeguards that can prevent some frustration for users of the widgets.

PR #158, opened 2017-05-09
Utilize wp.editor.remove()

westonruter:
See https://core.trac.wordpress.org/ticket/35760#comment:32

This also fixes a bug where the visual editor would fail to get re-switched to after dragging if the user had ever switched to the text tab.


timmyc:
Didn't see that `wp.editor.remove()` was new at first, after updating trunk this is working well in my testing.

timmyc (approved):
Looks good, nice refactor!

PR #164, opened 2017-05-09
Use 'Add X' instead of 'Select X' for buttons

### LINKED ISSUE #162: Change "Select [Media]" to "Add [Media]"
Not only it is more consistent with the "Add Media" button on posts/pages, it's also a little more clear that you can add new media from your library or from external sources, instead of just selecting a media file already in your library.

Props @michelleweber for the suggestion.

westonruter:
Fixes #162.


timmyc:
When testing this out, it looks like the media modal translation was missed:

Updated to use `add_media` in f251331

timmyc (approved):
If my commit looks good, then everything else tested out well for me.

westonruter:
Good catch!

Issue #128, opened 2017-05-04
The file types upload are not restricted according to the Widget type


miina:
It's possible to set an image to video/audio widget etc. Will be displayed as links / broken image.
![screen shot 2017-05-04 at 3 48 19 pm](https://cloud.githubusercontent.com/assets/3294597/25719834/2682d748-30e1-11e7-820d-b970646ea1c7.png)

westonruter:
To clarify: this is during the _upload_ flow. The media browser does successfully restrict the file types, but the _upload_ UI currently is not.

melchoyce:
Is this fixable?

timmyc:
I suppose we could do some extension testing in the selection callback to ensure we have a valid extension per the widget.  Kind of _feels_ like the media modal should prevent this though.

westonruter:
There should be a parameter to pass to the upload state to restrict the type.

westonruter:
Hummm. This seems like it may actually be a core bug. For example, if you go into edit a post, and you open the media modal to select a featured image. From this screen, you can go to upload a video and then _select_ it. That shouldn't be possible.

Presskopp:
FYI, there's a related ticket:
https://core.trac.wordpress.org/ticket/39625

westonruter:
I can see the full list of MIME types when I look at `wp.media.frame.uploader.uploader.uploader.settings.filters.mime_types`. I'm not sure how to properly override it, however.

westonruter:
I'm not the only one to ask the question: http://stackoverflow.com/questions/38400352/plupload-change-mime-type-filter-after-init

westonruter:
Looks like there is a bug in Plupload affecting the ability to change the file filter after init: https://github.com/moxiecode/plupload/issues/1339

joemcgill:
I've been digging into this issue a bit but haven't had time to code a solution yet. However, thought it might help to leave some initial thoughts.

[...]

PR #172, opened 2017-05-10
Show error when unsupported video type is selected.

timmyc:
This branch seeks to inform users that certain video file types are not supported via the `renderPreview()` in the widget itself.  The `.mov` issue detailed in #171 is a bit complex - and until uploads can be properly restricted ( #128 ) - users will still be able to upload a .mov file which will not function properly when the widget is rendered on the front-end of the site.

So this branch is implementing @jasmussen 's [idea](https://github.com/xwp/wp-core-media-widgets/issues/171#issuecomment-300549660) - to not render a functioning `<video />` preview in the widget for a file type that we know will not work.  I borrowed @obenland's error message suggestion:

__To Test__
- Create a video widget using a supported file type ( .webm, .m4v etc ) verify that the preview renders as expected
- Create a video widget using a .mov file, and verify the preview renders with an error message

### DIFF
--- a/wp-includes/widgets/class-wp-widget-media-video.php
+++ b/wp-includes/widgets/class-wp-widget-media-video.php
@@ -42,7 +42,7 @@ class WP_Widget_Media_Video extends WP_Widget_Media {
 			'media_library_state_multi' => _n_noop( 'Video Widget (%d)', 'Video Widget (%d)' ),
 			'media_library_state_single' => __( 'Video Widget' ),
 			/* translators: placeholder is a list of valid video file extensions */
-			'unsupported_file_type' => sprintf( __( 'Sorry, we can\'t display the video file type selected. Please upload a file with one of these extensions instead: %1$s' ), '<code>.' . implode( '</code>, <code>.', wp_get_video_extensions() ) . '</code>' ),
+			'unsupported_file_type' => sprintf( __( 'Sorry, we can&#8217;t display the video file type selected. Please upload a file with one of these extensions instead: %1$s' ), '<code>.' . implode( '</code>, <code>.', wp_get_video_extensions() ) . '</code>' ),
 		) );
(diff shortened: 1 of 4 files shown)


melchoyce:
Looks good 👍 

Can we tighten up the padding inside the notice code blocks to `padding: 0 3px 0 0px;`?

screen shot 2017-05-10 at 4 58 13 pm

Should "movie" be "video?"

timmyc:
Thanks for testing it out @melchoyce - addressed the verbiage and padding in 23eaa1b

westonruter (approved):

wp-includes/widgets/class-wp-widget-media-video.php
```diff
@@ -41,6 +41,8 @@ public function __construct() {
 			/* translators: %d is widget count */
 			'media_library_state_multi' => _n_noop( 'Video Widget (%d)', 'Video Widget (%d)' ),
 			'media_library_state_single' => __( 'Video Widget' ),
+			/* translators: placeholder is a list of valid video file extensions */
+			'unsupported_file_type' => sprintf( __( 'Sorry, we can\'t display the video file type selected. Please upload a file with one of these extensions instead: %1$s' ), '<code>.' . implode( '</code>, <code>.', wp_get_video_extensions() ) . '</code>' ),
```
obenland: `can\'t` should be `can&#8217;t`, I cut corners in my example

obenland (changes requested):

wp-includes/widgets/class-wp-widget-media-video.php
timmyc: Changed it up in 76783bc

obenland (approved):

jasmussen:
💖

PR #173, opened 2017-05-10
Limit the MIME types to just those that are embeddable

westonruter:
This prevents MOV files from appearing in the list files available to embed for the video widget. It won't currently prevent a MOV from being uploaded (#128) but it _does_ make it so that once you've uploaded a MOV it won't appear in the list, even though it _will_ be selected and appear in the attachment details with the Add to Widget button being enabled.

See #171.


timmyc (approved):
Verified that .mov files are no longer shown in the modal, nice work!

PR #175, opened 2017-05-11
Reduce padding inside code elements inside error notices

westonruter:
See https://github.com/xwp/wp-core-media-widgets/issues/149#issuecomment-300644019

Results:

image

image


timmyc:
Looks good to me, but will let @melchoyce give it the final 👍

melchoyce:
:shipit:

PR #176, opened 2017-05-11
Use new oembed/proxy endpoint instead of noembed

timmyc:
This branch removes usage of noembed and implements the new `oembed/proxy` REST API endpoint instead.  Requires to have the branch https://github.com/xwp/wordpress-develop/pull/225 installed to test.

### DIFF
--- a/wp-admin/js/widgets/media-video-widget.js
+++ b/wp-admin/js/widgets/media-video-widget.js
@@ -111,15 +111,16 @@
 			}
 
 			control.fetchEmbedDfd = jQuery.ajax({
-				url: 'https://noembed.com/embed',
+				url: wp.media.view.settings.oEmbedProxyUrl,
 				data: {
 					url: control.model.get( 'url' ),
 					maxwidth: control.model.get( 'width' ),
-					maxheight: control.model.get( 'height' )
+					maxheight: control.model.get( 'height' ),
+					_wpnonce: wp.media.view.settings.nonce.wpRestApi
 				},
 				type: 'GET',
-				crossDomain: true,
-				dataType: 'json'
+				dataType: 'json',
+				context: control
 			});
 
 			control.fetchEmbedDfd.done( function( response ) {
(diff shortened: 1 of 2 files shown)


wp-admin/js/widgets/media-widgets.js
```diff
@@ -182,15 +182,16 @@ wp.mediaWidgets = ( function( $ ) {
...
 
 						embedLinkView.dfd = $.ajax({
-							url: 'https://noembed.com/embed', // @todo Replace with core proxy endpoint once committed.
+							url: wp.media.view.settings.oEmbedProxyUrl,
 							data: {
 								url: embedLinkView.model.get( 'url' ),
 								maxwidth: embedLinkView.model.get( 'width' ),
-								maxheight: embedLinkView.model.get( 'height' )
+								maxheight: embedLinkView.model.get( 'height' ),
+								_wpnonce: wp.media.view.settings.nonce.wpRestApi
```
westonruter: Since we know that we only will be wanting YouTube and Vimeo, how about adding `discover=false` param?

wp-admin/js/widgets/media-video-widget.js
```diff
@@ -111,15 +111,16 @@
...
 
 			control.fetchEmbedDfd = jQuery.ajax({
-				url: 'https://noembed.com/embed',
+				url: wp.media.view.settings.oEmbedProxyUrl,
 				data: {
 					url: control.model.get( 'url' ),
 					maxwidth: control.model.get( 'width' ),
-					maxheight: control.model.get( 'height' )
+					maxheight: control.model.get( 'height' ),
+					_wpnonce: wp.media.view.settings.nonce.wpRestApi
```
westonruter: Also per below, add `discover=false`?

timmyc:
Good catch on the discover param, added in 4b8a03d

westonruter:
Committed to `trunk` in https://core.trac.wordpress.org/changeset/40641

I think we'll leave noembed.com as-is in the plugin for users on WP 4.7.

PR #179, opened 2017-05-11
Skip adding visual text widget if already merged into core

westonruter:
Useful for once https://github.com/xwp/wordpress-develop/pull/230 is committed to `trunk`.


obenland:
I'm not even sure this is technically needed. The only repercussion of running both simultaneously is  that `wp.textWidgets.init()` gets called twice :)

obenland (approved):
:shipit:

westonruter:
Actually I think the problem is the `text-widgets` JS file from core continues to get used whereas the PHP class from the plugin then gets used, and things can get inconsistent. So better to just make sure only one version is loaded at a time.

PR #181, opened 2017-05-11
Fix race condition if hitting Save & Publish button too quickly after entering text

westonruter:
See https://core.trac.wordpress.org/ticket/40729

### DIFF
--- a/wp-admin/js/widgets/text-widgets.js
+++ b/wp-admin/js/widgets/text-widgets.js
@@ -162,7 +162,28 @@ wp.textWidgets = ( function( $ ) {
 
 				control.editorFocused = false;
 				triggerChangeIfDirty = function() {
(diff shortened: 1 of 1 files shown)


obenland:
I'm honestly not sure what exactly to look for here? It might also be that my local environment is to fast for me to run into that race condition?

westonruter:
@obenland sorry, here is a video depicting the problem that this PR fixes: https://cloudup.com/cv94KypYpkC

Notice in the video how I move my mouse over to the Save & Publish button while keyboard focus remains in the editor. I type “3” into the editor and then immediately (within 250ms) hit Save & Publish. You'll note that the button becomes re-enabled after saving and inspecting `wp.customize.dirtyValues()` will show that the widget is still dirty. This is because without the change in this PR, the save operation isn't properly blocked while waiting for the `update-widget` request to initialize.

obenland (approved):
Yes, I was able to reproduce and patch fixes it

PR #187, opened 2017-09-07
Remove attachment data from gallery widget schema.

joemcgill:
To avoid saving attachment data with the widget, this removes attachments
from the schema and makes use of the IDs we're saving to fetch attachment
data before the widget preview is renedered.

### DIFF
--- a/wp-admin/js/widgets/media-gallery-widget.js
+++ b/wp-admin/js/widgets/media-gallery-widget.js
@@ -93,7 +93,7 @@
 			previewContainer = control.$el.find( '.media-widget-preview' );
 			previewTemplate = wp.template( 'wp-media-widget-gallery-preview' );
 
-			if ( control.model.get( 'ids').length && ! control.model.get( 'attachemnts' ) ) {
+			if ( control.model.get( 'ids' ).length && ! control.model.get( 'attachemnts' ) ) {
 				attachments = this.getAttachments();
 
 				attachments.more().done( function() {
@@ -120,7 +120,7 @@
 				perPage: -1,
 				post__in: ids,
 				query: true,
-				type: 'image',
+				type: 'image'
 			} );
 
 			return attachments;
--- a/wp-admin/js/widgets/media-gallery-widget.js
+++ b/wp-admin/js/widgets/media-gallery-widget.js
@@ -88,11 +88,44 @@
 		 * @returns {void}
 		 */
 		renderPreview: function renderPreview() {
-			var control = this, previewContainer, previewTemplate;
+			var control = this, previewContainer, previewTemplate, attachments;
+
 			previewContainer = control.$el.find( '.media-widget-preview' );
 			previewTemplate = wp.template( 'wp-media-widget-gallery-preview' );
-			previewContainer.html( previewTemplate( control.previewTemplateProps.toJSON() ) );
+
+			if ( control.model.get( 'ids').length && ! control.model.get( 'attachemnts' ) ) {
+				attachments = this.getAttachments();
+
+				attachments.more().done( function() {
+					control.model.set( 'attachments', JSON.stringify( _.pluck( attachments.models, 'attributes' ) ) );
+					previewContainer.html( previewTemplate( control.model.attributes ) );
+				} );
+			} else {
+				previewContainer.html( previewTemplate( control.previewTemplateProps.toJSON() ) );
+			}
 		},
+
+		/**
+		 * Fetch attachment models.
+		 *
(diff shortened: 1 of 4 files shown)


joemcgill:
Didn't want to stomp on #120 directly, so opening a new PR against the `add/gallery-widget` branch.

wp-includes/widgets/class-wp-widget-media-gallery.php
```diff
@@ -89,10 +89,6 @@ public function get_instance_schema() {
 				'media_prop'            => '_orderbyRandom',
 				'should_preview_update' => false,
 			),
-			'attachments' => array(
-				'type'                  => 'string',
-				'default' => '',
-			),
```
westonruter, line 95: Shouldn't this be retained, but have an `array` type with a default of `array()`?

wp-admin/js/widgets/media-gallery-widget.js
```diff
@@ -88,11 +88,44 @@
...
 		},
+
+		/**
+		 * Fetch attachment models.
+		 *
+		 * @returns {wp.media.model.Attachments} A Backbone.Collection.
+		 */
+		getAttachments: function getAttachments() {
+			var attachments,
+				ids = this.model.get( 'ids' ).split( ',' );
```
westonruter, line 115: The underlying widget instance can store an array instead of a string.

wp-includes/widgets/class-wp-widget-media-gallery.php
joemcgill: I'm new to these media widgets, but I don't think we need attachments in the schema at all unless there's some way to keep particular schema from being turned into input fields which get saved to the widget data (which is what we're trying to avoid).

wp-admin/js/widgets/media-gallery-widget.js
joemcgill: 👍 Will play with that idea.

joemcgill: Actually, if we were to do that, then we'll have to convert the array to a string on the front end before rendering the shortcode. Not sure which is preferred?

westonruter: @joemcgill you mean when `ids` is passed into `gallery_shortcode()`?  I can see that this will get mapped to the `include` arg which is passed to `get_children()` and then down into `get_posts()` which will then pass it into `wp_parse_id_list( $r['include'] )`, which accepts both an array and a string. So I don't think it will get converted to a string?

wp-includes/widgets/class-wp-widget-media-gallery.php
westonruter: The schema is used by the widget's `update` method to handle sanitization and validation of the instance properties: https://github.com/xwp/wp-core-media-widgets/blob/1089ca34acba917e303f4b8a72e72284d63135cb/wp-includes/widgets/class-wp-widget-media.php#L247-L272

westonruter: And actually, the property could be:

```php
'attachments' => array(
    'type' => 'array',
    'default' => array(),
    'items' => 'integer',
),
```

To make it an explicit array of integer values.

westonruter: Aside: the Travis build is failing because `attachments` is not in the schema: https://travis-ci.org/xwp/wp-core-media-widgets/jobs/272758064#L650

joemcgill: Right. We're already saving `ids` as part of the schema, so unless I'm missing something, I think we really do want to remove attachments from the schema, because we're not saving any attachment data, instead, we're generating the attachment data in the preview only.

westonruter: Oh 🤦‍♂️ you're right! Sorry, I missed that.

So yes, you're right to remove `attachments`. So then I think it is the `ids` that can be changed to have an `array` type if we want:

```php
'ids' => array(
    'type' => 'array',
    'default' => array(),
    'items' => 'integer',
),
```

joemcgill: Currently, it looks like schema doesn't support an array type, because we would need to have some way to transform an array of data to/from the hidden input field that is generated for the widget form.

westonruter: Good point. The widget field syncing is assuming scalar values for the model. Alas, someday we'll get proper models for ~widgets~blocks.

PR #188, opened 2017-09-08
Update the gallery widget preview design.

joemcgill:
This is an initial implementation of the enhanced widget preview
design proposed in #62.

See: https://cloud.githubusercontent.com/assets/2846578/25029086/e8b7efa2-2087-11e7-8ea6-43679a3a2fb8.png

### DIFF
--- a/core-media-widgets.php
+++ b/core-media-widgets.php
@@ -24,7 +24,8 @@
  * @package WordPress
  */
 
-define( 'WP_CORE_MEDIA_WIDGETS_MERGED', file_exists( ABSPATH . 'wp-includes/widgets/class-wp-widget-media.php' ) );
+define( 'WP_CORE_MEDIA_WIDGETS_MERGED',      file_exists( ABSPATH . 'wp-includes/widgets/class-wp-widget-media.php' ) );
+define( 'WP_CORE_GALLERY_WIDGET_MERGED',     file_exists( ABSPATH . 'wp-includes/widgets/class-wp-widget-media-gallery.php' ) );
 define( 'WP_CORE_VISUAL_TEXT_WIDGET_MERGED', file_exists( ABSPATH . 'wp-admin/js/widgets/text-widgets.js' ) );
 
 // Register WP-CLI command for generating QUnit test suite.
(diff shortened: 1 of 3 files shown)


joemcgill:
A couple of questions about this current implementation:

In the designs, @melchoyce specified showing the number of additional images with placeholder text reading "+ %d more", but that number is only available inside the template. I'm not exactly sure the best way to combine that data with a translatable string, so suggestions welcome.

There's probably a better way to enqueue the CSS here, but I couldn't get a callback on `wp_default_styles` to work.

westonruter:
> showing the number of additional images with placeholder text reading "+ %d more", but that number is only available inside the template. I'm not exactly sure the best way to combine that data with a translatable string, so suggestions welcome.

That is currently a challenge in JS, since there isn't yet i18n in core ([#20491](https://core.trac.wordpress.org/ticket/20491)). There was a [workaround](https://github.com/WordPress/wordpress-develop/commit/1e7b15752336580a0d9e16a7333f47fc71f7854d) we did in [#39087](https://core.trac.wordpress.org/ticket/39087), but it wasn't pretty as we basically had to iterate over all possible numbers on the server and then export them to JS:

https://github.com/WordPress/wordpress-develop/blob/f5c342ce7671731dc24bc2e228640bdae767e87b/src/wp-includes/class-wp-customize-widgets.php#L707-L746

So that is one option. Maybe strings for numbers 1 to 25 could be generated. Or maybe the string could avoid needing to be translated by just omitting “more”? Or rather, the translated string could be just `__( '+%d' )` and then it would be up to translators to modify this string to an appropriate math-like expression in the target language, but being math hopefully it wouldn't even need to be translated in the first place.

westonruter (approved):
Only other things I suggest would be to let the label be “+X” instead of “+More”. Also, clicking on the gallery preview I think should be a shortcut for “Edit Gallery”.

joemcgill (commented):
I swear I tried this and couldn't get it to work, but 🤷‍♂️. Seems good now.

westonruter:
@joemcgill haha! Behold what I just noticed when creating a Facebook post: `+4`

image

joemcgill:
@westonruter I updated the template to show the number instead of "More" until we have a better way of handling i18n in JS.

Preview gallery image

However, I _do_ like the idea of putting that over an actual image instead of a grey box.

melchoyce:
Let's try the +# over an image with a 75% black overlay (or 75% opacity over a black bg, whichever) with white text, and bump up the font size a little.

joemcgill:
How do you like this?

text over image

westonruter:
Looks amazing! (Ignore that Travis build error. It's a failure not an error on your part.)

melchoyce:
I think we can actually go lighter on the opacity, but let's merge and then fiddle later.

